PR #117, opened 2021-04-26
536. Construct Binary Tree from String (Recursive solution)

altayhunoglu:
Please refer to https://github.com/spiralgo/algorithms/issues/49 for a detailed explanation.

We have three behavioral cases for the current character at any traverse: '(', ')' and a digit.
(and one extra character for the '-' sign, if the digit is negative.)
We use a switch-case statement to determine behaviour for each case.

'(' means we need a new node. 
')' means we have to turn back to the root.
Any digit: means we have to set the current root's value with that digit.
image

### DIFF
--- a/src/main/java/algorithms/curated170/medium/ConstructBinaryTreefromStringRecursive.java
+++ b/src/main/java/algorithms/curated170/medium/ConstructBinaryTreefromStringRecursive.java
@@ -1,0 +1,68 @@
+package algorithms.curated170.medium;
+
+import algorithms.datastructures.TreeNode;
+
+public class ConstructBinaryTreefromStringRecursive {
+
+    public static void main(String[] args) {
+        String s = "-60(9(8)(7))(5)";
+        ConstructBinaryTreefromStringRecursive cbts = new ConstructBinaryTreefromStringRecursive();
+        TreeNode previous = cbts.str2tree(s);
+        System.out.println();
+    }
+
+    String globalString;
+    int len;
+    int i = 0;
+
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/ConstructBinaryTreefromStringRecursive.java
```diff
@@ -0,0 +1,68 @@
...
+                    i++;
+                    sign = -1;
+                }
+
+                int num = 0;
+                while (i < len && Character.isDigit(globalString.charAt(i))) {
+                    num = num * 10 + globalString.charAt(i++) - '0';
+                }
+
+                root = new TreeNode(num * sign);
```
ErdemT09, line 61: Instead of,
`                    num = num * 10 + globalString.charAt(i++) - '0';
`
we can use Java's built-in integer parser if we want. 
`                    num = num * 10 + Character.getNumericValue(globalString.charAt(i++));
`

altayhunoglu: Yeah but this method is faster then the built-in one.

ErdemT09 (approved):
Such questions relating to parsing a string to some kind of a data type or statement is an important pattern and I believe is related to the functionality of compilers. Doing this through a recursive, DFS-like way is appropriate as we have nested parentheses. That is the second pattern we can observe in this solution. If the question had included inputs with empty parentheses, this might not have worked but the question lets assume that these are not the case. Approved.

PR #118, opened 2021-04-27
NestedListWeightSum2 with HashMap

ErdemT09:
In this solution, in addition to keeping track of the maximum depth of the nested integers, we also sum individual layers and store these in a map. At the end, instead of re-looping through the nested integers, we just add the layer-sums elements in the map with their distance from the maximum depth.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/NestedListWeightSum2HashMap.java
+++ b/src/main/java/algorithms/curated170/medium/NestedListWeightSum2HashMap.java
@@ -1,0 +1,37 @@
+package algorithms.curated170.medium;
+
+import java.util.HashMap;
+import java.util.List;
+
+class NestedIntegerWeightSum2Hashmap {
+    HashMap<Integer, Integer> deepSumMap = new HashMap<>();
+    int max = 1;
+
+    public int depthSumInverse(List<NestedInteger> nestedList) {
+        maxDepth(nestedList, 1);
+        return sumMapTotal();
+    }
+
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/NestedListWeightSum2HashMap.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public int depthSumInverse(List<NestedInteger> nestedList) {
+        maxDepth(nestedList, 1);
+        return sumMapTotal();
+    }
+
+    private void maxDepth(List<NestedInteger> nInt, int depth) {
+        for (NestedInteger nestedInteger : nInt) {
+            if (nestedInteger.isInteger()) {
+                deepSumMap.put(depth, deepSumMap.get(depth)+nestedInteger.getInteger());
```
altayhunoglu, line 18: Can we alter this for the null-check:
                deepSumMap.put(depth, deepSumMap.getOrDefault(depth, 0)+nestedInteger.getInteger());

ErdemT09: Yes, null-checks or other checks like these shouldn't be forgotten.

ErdemT09: Added.

altayhunoglu (approved):
The code is quite clear and smart. Thanks.

PR #119, opened 2021-04-27
Sum All Integers in a String

ErdemT09:
Quesstion: Given a string str, return the sum of integers in this string (sign is also considered).

I used here the algorithm for parsing integers to string that we had used in [536. Construct Binary Tree from String](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/medium/ConstructBinaryTreefromStringRecursive.java) and  [604. Design Compressed String Iterator](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/easy/DesignCompressedStringIterator.java). This problem and the solution here is another demonstration of this algorithm, good as a simpler example.

### DIFF
--- a/src/main/java/algorithms/SumAllIntegersInString.java
+++ b/src/main/java/algorithms/SumAllIntegersInString.java
@@ -8,16 +8,15 @@ public class SumAllIntegersInString {
         for (int i = 0; i < str.length(); i++) {
             char ch = str.charAt(i);
             if (ch == '-') {
-                if (currNum != 0) {
-                    sum += currNum * sign;
-                    currNum = 0;
-                }
+                sum += currNum * sign;
+                currNum = 0;
+
                 sign = -1;
                 continue;
             }
             if (Character.isDigit(ch)) {
                 currNum = currNum * 10 + ch - '0';
-                //currNum = currNum * 10 + Character.getNumericValue(ch);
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/SumAllIntegersInString.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class SumAllIntegersInString {
+    public static int sumAllIntsInString0(String str) {
+        int sum = 0;
+        int currNum = 0;
+        int sign = 1;
+        for (int i = 0; i < str.length(); i++) {
+            char ch = str.charAt(i);
+            if (ch == '-') {
+                if (currNum != 0) {
```
altayhunoglu, line 11: We may not need this if-statement in this case.

ErdemT09: By line 11 or line 10? If it's line 11, we can indeed get rid of it.

altayhunoglu: Yeah, it is 11; as, even if 'currNum' is '0' it will not affect.

ErdemT09: Updated.

altayhunoglu (approved):
Quite clean code. Thanks.

PR #122, opened 2021-04-28
1102. Path With Maximum Minimum Value 

altayhunoglu:
It is possible to solve this question using PriorityQueue, Union Find, Binary Search.

 Binary Search looks like the fastest one and it is based on the idea that the value should be in a range between 0 and (high = Math.min(A[0][0], A[row-1][col-1])) as it is certain that any possible path must include both of these nodes.
  
  Refer to:https://github.com/spiralgo/algorithms/issues/50 for a detailed explanation (currently in Turkish)
  
image

### DIFF
--- a/src/main/java/algorithms/curated170/medium/PathWithMaximumMinimumValue.java
+++ b/src/main/java/algorithms/curated170/medium/PathWithMaximumMinimumValue.java
@@ -1,0 +1,54 @@
+package algorithms.curated170.medium;
+
+public class PathWithMaximumMinimumValue {
+
+    public static void main(String[] args) {
+        PathWithMaximumMinimumValue pathWithMaximumMinimumValue = new PathWithMaximumMinimumValue();
+
+        int[][] A = {{5, 4, 5},
+        {1, 2, 6},
+        {7, 4, 6}};
+
+        int result = pathWithMaximumMinimumValue.maximumMinimumPath(A);
+        System.out.println(result);
+    }
+    final static int DIRECTIONS[] = {0, 1, 0, -1, 0};
+
+    public int maximumMinimumPath(int[][] A) {
+        int rowCount = A.length, colCount = A[0].length;
+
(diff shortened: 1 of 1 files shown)


ErdemT09:
n ve m grid boyutları diyelim. A ise 0,0 noktasının ve bitiş noktasının değerlerinin en azı olsun (high değişkeninin bağlı olduğu değer).
Bunun zaman karmaşıklığı o zaman O(m*n*log(A)) mı oluyor? Her bölerek aramada tüm gridi dolanıyoruz. Bunu log(A)'ya bağlı bir biçimde yapacağımız böyle düşündüm.

altayhunoglu:
Evet, haklısın.
Soruda şöyle verilmiş:
0 <= A[i][j] <= 10^9

Demek ki, Binary Search en azından şöyle arayacak:
O(log n) = O(log 10^9).
2 tabanında log 10^9= 29.897352854 ediyor. Yani, yaklaşık 30 adımda, halletmiş olacak.
isThereAPossiblePath için de muhtemelen n*m gibi bir yük eklenir.
Dolayısı ile, doğru söylüyorsun.

altayhunoglu:
You are absolutely right. I refactored and pushed again.

ErdemT09:
To consider the patterns from this answer, I would say that the method isThereAPossiblePath utilizes a kind of DFS. This kind of grid traversal with the boolean matrix keeping visited nodes is also notable. Noticing that the wanted return value is something 0 and the minimum value of the both ends of the grid is the another key point in this answer.
Besides these, showing directions like `final static int DIRECTIONS[] = {0, 1, 0, -1, 0};` was unreadable at first. But then I have noticed how it works and is shorter than writing all the 4 directions. Some small, but smart detail here.
So, I give an approval to this commit. However, I think that we should eventually at some point add the queue and the union set solutions to this question as wellç

ErdemT09 (approved):

PR #125, opened 2021-04-29
616. Add Bold Tag in String, using a boolean array.

altayhunoglu:
Refer to: https://github.com/spiralgo/algorithms/issues/72

 In the first place, we mark all characters which will be bold.
 We use a boolean array to mark the characters of the string (**s**).
 A character should be marked if it is a char of any substring in the dictionary (**dict**).

(Note that, although the dictionary does not contain duplicates, any substring in the dictionary may occur multiple times in the string 's' )

After the marking process, it is quite easy to wrap marked parts of the string using a StringBuilder.

image

### DIFF
--- a/src/main/java/algorithms/curated170/medium/AddBoldTaginString.java
+++ b/src/main/java/algorithms/curated170/medium/AddBoldTaginString.java
@@ -2,7 +2,7 @@ package algorithms.curated170.medium;
 
 public class AddBoldTaginString {
 
-    public String addBoldTag(String s, String[] dict) {
+ public String addBoldTag(String s, String[] dict) {
         StringBuilder sb = new StringBuilder();
         boolean arr[] = new boolean[s.length()];
         for(String d : dict){
@@ -17,17 +17,14 @@ public class AddBoldTaginString {
         }
         boolean startB = false;
         for(int i = 0; i < arr.length; i++){
-            if(arr[i]){
-                if(!startB){
+            if(arr[i] && !startB){
                     startB = true;
                     sb.append("<b>");
-                }
-            }else{
-                if(startB){
+             }else if(!arr[i] && startB){
                     startB = false;
                     sb.append("</b>");
                 }
-            }
+            
             sb.append(s.charAt(i));
         }
         if(startB){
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/AddBoldTaginString.java
```diff
@@ -0,0 +1,38 @@
...
+                if(!startB){
+                    startB = true;
+                    sb.append("<b>");
+                }
+            }else{
+                if(startB){
+                    startB = false;
+                    sb.append("</b>");
+                }
+            }
```
ErdemT09, line 30: These are equivalent to 
```
if(arr[i] && !startB){
                    startB = true;
                    sb.append("<b>");
            }

else if(startB){
                    startB = false;
                    sb.append("</b>");
            }
```

src/main/java/algorithms/curated170/medium/AddBoldTaginString.java
```diff
@@ -0,0 +1,38 @@
...
+        StringBuilder sb = new StringBuilder();
+        boolean arr[] = new boolean[s.length()];
+        for(String d : dict){
+            int start = s.indexOf(d);
+            while(start != -1){
+                int length = start+ d.length();
+                for(int i = start; i < length; i++){
+                    arr[i] = true;
+                }
+                start = s.indexOf(d, start + 1);
```
ErdemT09, line 15: Is this used to check the recurrences of the same string `d` inside of the string `s` after its previously detected index?

src/main/java/algorithms/curated170/medium/AddBoldTaginString.java
altayhunoglu: Yeah, you are right. I just thought the first one was more readable. But, we can alter the code. No problem.

altayhunoglu: Oke, we fixed it like this.

```
  if(arr[i] && !startB){
                    startB = true;
                    sb.append("<b>");
             }else if(!arr[i] && startB){
                    startB = false;
                    sb.append("</b>");
                }
            
```

ErdemT09 (approved):
Notable for its usage of StringBuilder and its methods.  Using a boolean array to keep track of some values reminded me of [Counting Primes](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/easy/CountPrimes.java) and [Subset Sum ](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/SubsetSum.java#L48).

PR #126, opened 2021-04-29
Longest Substring with At Most K Distinct Characters 

altayhunoglu:
(Refer to:#71)

We employed the "Sliding Window Technique" using two pointers (left, right).

While we are sliding the window till the end of the given string; we check if we exceed the limit that is represented with "k".
Until we satisfy the condition (distinct chars count < k) we shrink the string from start. In this way, we can produce all "possible lengths" with at most k distinct characters.

In the end, we get the maximum of all "possible lengths" and return the result.

In the first place, using HashMap for both storing data and calculating "distinct " using the map's size.
But merge and remove operations are more efficient in the array.
Nevertheless, I add the HashMap solution along with the one using an array.

116593836-3520e180-a92a-11eb-82b4-6374f9d400ba

### DIFF
--- a/src/main/java/algorithms/curated170/medium/LongestSubstringwithAtMostKDistinctCharacters.java
+++ b/src/main/java/algorithms/curated170/medium/LongestSubstringwithAtMostKDistinctCharacters.java
@@ -1,0 +1,40 @@
+package algorithms.curated170.medium;
+  
+
+public class LongestSubstringwithAtMostKDistinctCharacters {
+
+    public static void main(String[] args) {
+    
+        LongestSubstringwithAtMostKDistinctCharacters lo = new LongestSubstringwithAtMostKDistinctCharacters();
+        int output = lo.lengthOfLongestSubstringKDistinct("abbccc", 2);
+        System.out.println(output);
+    }
+
+    public int lengthOfLongestSubstringKDistinct(String s, int k) {
+       if(k >= s.length()) return s.length();
+      
+        char[] counterMap = new char[256];
+        char[] array = s.toCharArray();
+        int left = 0, right = 0;
+        int currentLongest = 0;
+        int distinct = 0;
+         
+        for(int ch: array){
+            if(counterMap[ch] == 0) distinct++;
+            counterMap[ch]++;
+            
+            while(distinct>k){
+                 char currShrink = array[left++];
+                 if(counterMap[currShrink]==1) distinct--;
+                 counterMap[currShrink]--;
+                  
+            }
+          
+           
+            currentLongest = Math.max(currentLongest, right - left + 1);
+            right++;
+        }
+
+        return currentLongest;
+    }
+}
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/medium/LongestSubstringwithAtMostKDistinctCharacters.java
```diff
@@ -0,0 +1,40 @@
...
+    
+        LongestSubstringwithAtMostKDistinctCharacters lo = new LongestSubstringwithAtMostKDistinctCharacters();
+        int output = lo.lengthOfLongestSubstringKDistinct("abbccc", 2);
+        System.out.println(output);
+    }
+
+    public int lengthOfLongestSubstringKDistinct(String s, int k) {
+       if(k >= s.length()) return s.length();
+      
+        char[] counterMap = new char[256];
```
ErdemT09, line 16: I am not sure, but this might also be possible to implement with a boolean array.

ErdemT09: Şöyle bir şey, eğer bir char'la karşılaşmamışsak onu görüp kalan değişik harf sayısını azaltıyoruz. 
```
public static int solution(String s, int k)
    {
        int res = 0;
        for (int i = 0; i < s.length(); i++) {
            int r = i+1;
            boolean[] hasLetter = new boolean[256];
            hasLetter[s.charAt(i)] = true;
            int dis = k;
            while(r<s.length() && dis>0)
            {
                if(!hasLetter[s.charAt(r)])
                {
                    dis--;
                    hasLetter[s.charAt(r)] = true;
                }
                r++;
            }
            //System.out.println(i + " " + r);
            r = r==s.length() ? r : r-1;
            res = Math.max(res, r-i);
        }
        return res;
    }
```

ErdemT09: Bu i'yi dinamik olarak taşıyarak optimize edilebilir.

ErdemT09: It's actually quite inefficient. It can be considered a naive solution.

ErdemT09, line 16: With further consideration, this is actually a char array that is used to hold information about numbers, namely, how many times a give char at index of its ASCII value occurs. Sort of like a "unsigned short".

ErdemT09: It's sort of unusual, as in normal char arrays: key = int and value = char,
but here: key = char and value = int.

ErdemT09 (approved):
This question is a good example as a question using two pointers.

PR #127, opened 2021-04-29
159. Longest Substring with At Most Two Distinct Characters

altayhunoglu:
It is just a simpler instance of that problem:
https://github.com/spiralgo/algorithms/issues/71

So it will be working as well, after merging that branch in the PR:
https://github.com/spiralgo/algorithms/pull/126

image


ErdemT09 (approved):
Basically the simpler version of the `Longest Substring with At Most K Distinct Characters`, same logic is valid here.

PR #128, opened 2021-04-30
360. Sort Transformed Array

altayhunoglu:
(Refer to https://github.com/spiralgo/algorithms/issues/52 for detailed explanation)

It is possible to solve this question using PriorityQueue or built-in sort functions. 
Although I add the naive solution with PriorityQueue; it would be great to have some mathematical sense on the occasion of this great question:

A quadratic function of the form _**f(x) = ax2 + bx + c**_ is given.
The graph of a quadratic function called a **parabola**.

We have two possible cases (and one more insignificant case a == 0 as it will not affect the outcome.)
parabola

**Case 1:**
If a > 0 (positive) then the parabola opens upward.
 In this case, the middle "y" will be the minimum and one of the two ends (y1, y2) will be the maximum. 
 As y1, y2 must be greater than others; we need to compare these y pairs at each iteration and insert them into the end of the array.

 **Case 2:**
If a < 0 (negative) then the parabola opens downward.

 In this case, the middle "y" will be the maximum and one of the two ends (y1, y2) will be the minimum. 
 As y1, y2 must be smaller than others; we need to compare these y pairs at each iteration and insert them into the start of the array.

image

### DIFF
--- a/src/main/java/algorithms/curated170/medium/SortTransformedArray.java
+++ b/src/main/java/algorithms/curated170/medium/SortTransformedArray.java
@@ -1,0 +1,45 @@
+package algorithms.curated170.medium;
+
+public class SortTransformedArray {
+
+    public class Solution {
+
+        public int[] sortTransformedArray(int[] nums, int a, int b, int c) {
+            int n = nums.length, start = 0, end = n - 1;
+            int[] res = new int[n];
+            
+            if (a > 0) {
+                for (int i = n - 1; i >= 0; i--) {
+                    int y1 = quadratic(nums[start], a, b, c);
+                    int y2 = quadratic(nums[end], a, b, c);
+                    if (y1 > y2) {
+                        res[i] = y1;
+                        start++;
+                    } else {
+                        res[i] = y2;
+                        end--;
+                    }
+                }
+            } else {
+                for (int i = 0; i < n; i++) {
+                    int y1 = quadratic(nums[start], a, b, c);
+                    int y2 = quadratic(nums[end], a, b, c);
+                    if (y1 < y2) {
+                        res[i] = y1;
+                        start++;
+                    } else {
+                        res[i] = y2;
+                        end--;
+                    }
+                }
+            }
+
+            return res;
+        }
+
+        int quadratic(int x, int a, int b, int c) {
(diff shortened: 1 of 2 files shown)


ErdemT09:
My idea was the reduce to reduce unnecessary parameters and reappearing code. I think this won't make any performance any performance differences but this looks more modular to me.
```
public class Solution {

        int n, start = 0, end;
        int[] res = new int[n];
        int[] nums;
        int a, b, c;

        public int[] sortTransformedArray(int[] nums, int a, int b, int c) {
            n = nums.length;
            end = n-1;
            this.nums = nums;
            this.a = a;
            this.b = b;
            this.c = c;
            
            if (a > 0) {
                for (int i = n - 1; i >= 0; i--) {
                    quarticComparator(i, true);
                }
            } else {
                for (int i = 0; i < n; i++) {
                    quarticComparator(i, false);
                }
            }

            return res;
        }

        void quarticComparator(int i, boolean pbType) {
            int y1 = quadratic(nums[start]);
            int y2 = quadratic(nums[end]);
            if ((pbType && y1 > y2) || (!pbType && y1 < y2)) {
                res[i] = y1;
                start++;
            } else {
                res[i] = y2;
                end--;
            }
        }

        int quadratic(int x) {
            return a * x * x + b * x + c;
        }
    }
```

altayhunoglu:
@ErdemT09 your refactoring is quite fine and reasonable.
It is why I'll be pushing back.

image

src/main/java/algorithms/curated170/medium/SortTransformedArray.java
```diff
@@ -0,0 +1,47 @@
...
+            res = new int[n];
+            end = n-1;
+            this.nums = nums;
+            this.a = a;
+            this.b = b;
+            this.c = c;
+            
+            if (a > 0) {
+                for (int i = n - 1; i >= 0; i--) {
+                    quarticComparator(i, true);
```
ErdemT09, line 21: This should be called quadraticComparator, quartic is something relating to the power of 4 (x^4).

ErdemT09 (approved):
This question is another example of a two pointers problem. It also shows how mathematics can come up as important in algorithms. I give my approval.

PR #132, opened 2021-05-01
First unique

altayhunoglu:
This question could be solved using many techniques and data structures.
 But the main difficulty of this question is on the constructor method.
 We have an integer array that can have elements up to 10^5; nevertheless, the unique element might be at the very end of this array.
 So, we have to read this array and somehow "mark" the unique ones without distorting the insertion order.
 We can use a LinkedHashMap (or two Queues, one for the unique elements) for this purpose as it maintains the insertion order.
 But, then again, it would be the naive solution. (It takes 142 ms on LeetCode.)
 Though I added this naive solution with LinkedHashMap as well.
 
 Instead of that, we can employ our own LinkedList implementation just we have done for [DesignALeaderboardFast](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/medium/DesignALeaderboardFast.java) as update-delete operations can be done more efficiently. (Even if it is faster than the naive solution, it takes 95 ms.)
 
 Our Node implementation should have a boolean variable: "active".
 Because we need to check if the map contains the element even if it is not unique.
 It is why instead of deleting the node directly, we just mark it as inactive.
 
 
 (Note: Another solution to this "constructor difficulty" could be marking the unique ones using a boolean array. But I did not find enough time to meditate on that solution.)
 https://github.com/spiralgo/algorithms/issues/53
 
image


ErdemT09:
Java does have a built-in linked hash set class. Perhaps we can use that and keep track of the all numbers added somewhere else.

```
import java.util.HashSet;
import java.util.LinkedHashSet;
import java.util.Set;

public class FirstUnique {

    HashSet<Integer> uniques = new LinkedHashSet<>();
    HashSet<Integer> allNums = new HashSet<>();

    public FirstUnique(int[] nums) {
        for (int k : nums) {
            add(k);
        }
    }

    public int showFirstUnique() {
        if (uniques.size() > 0) {
            return uniques.iterator().next();
        }
        return -1;
    }

    public void add(int k) {
        if (!allNums.add(k)) {
            uniques.remove(k);
        } else {
            uniques.add(k);
        }
    }

        FirstUnique fu = new FirstUnique(new int[] { 3, 2, 5, 5, 7, 11, -9, 5 });
        fu.add(2);
        fu.add(3);
        fu.add(3);
        fu.add(-4);
        fu.add(-1);
        System.out.println(fu.showFirstUnique()); // prints 7
    }
}
```

altayhunoglu:
> Java does have a built-in linked hash set class. Perhaps we can use that to check the queue and keep track of the all numbers added somewhere else.
 
 
 Yeah, of course, we should add this as another naive solution. But then again, it takes a long time because of the constructor.
 
 
image

ErdemT09:
That's a bit unexpected, as hash set should have O(1) removal and insertion.

altayhunoglu:
Maybe it is normal when we consider:
1 <= nums.length <= 10^5
1 <= nums[i] <= 10^8

ErdemT09 (approved):
Thank you for the good implementation of the node structure, this structure is noteworthy. Approved.

PR #142, opened 2021-05-02
1428. Leftmost Column with at Least a One

altayhunoglu:
(Refer to: https://github.com/spiralgo/algorithms/issues/54)

We can use either the row-first approach or the col-first approach for the solution.
Both would have almost the same efficiency. (0 ms.)
It is why I add both implementations.

For the row-first approach, we should consider that rows are sorted in a "non-decreasing" order.
It means we can employ Binary Search here.

For Binary Search we could look for _**"if (binaryMatrix.get(currentRow, currentCol) == 1)"**_  instead of _**if (binaryMatrix.get(currentRow, currentCol) == 0)**_
 But in this case, as far as I roughly think, we would have to use "Math.min()" in every iteration of while to reset currentResult until we find the smallest index.
 
For the col-first approach, we just increase the row number or decrease the column number (starting from the first row and the last column.) until we find the minimum index.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/BinaryMatrix.java
+++ b/src/main/java/algorithms/curated170/medium/BinaryMatrix.java
@@ -1,0 +1,12 @@
+package algorithms.curated170.medium;
+ 
+import java.util.List;
+
+/**
+ *
+ * @author Altay
+ */
+public interface BinaryMatrix {
+      public int get(int row, int col); 
+      public List<Integer> dimensions();
+}
--- a/src/main/java/algorithms/curated170/medium/BinaryMatrixImplement.java
+++ b/src/main/java/algorithms/curated170/medium/BinaryMatrixImplement.java
@@ -1,0 +1,24 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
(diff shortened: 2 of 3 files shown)


ErdemT09:
The constraint in the question is that we are not allowed to make more than a 1000 function calls.  The first solution clearly calls the get method m * log(n) times in the worst case. Since row and column can be at most 100 in this question, we never make more than 1000 method calls.
image

However, I'm not sure if the col-first one is linearithmic. It might actually be faster than that. Perhaps even O(n+m).

ErdemT09 (approved):
PR approved.

altayhunoglu:
> 100

Wonderful explanation, thanks.

altayhunoglu:
> The constraint in the question is that we are not allowed to make more than a 1000 function calls. The first solution clearly calls the get method m * log(n) times in the worst case. Since row and column can be at most 100 in this question, we never make more than 1000 method calls.
> image
> 
> However, I'm not sure if the col-first one is linearithmic. It might actually be faster than that. Perhaps even O(n+m).



Wonderful explanation. Thanks.

PR #165, opened 2021-05-04
A suggestion for how different answers to the same question should be located in the project

ErdemT09:
We basically put them into the same folder. With this, finding answers to questions are also easier. This could also be done for all individual questions, especially those with not-easy-to-search names. 
However, I'm not sure if this will properly function with java packages. That's why I'm skeptical of it.


altayhunoglu:
Brilliant idea Erdem.
Categorizing on the problem name is quite fine according to me.

Just as a note, we should consider this folder will be a Java package.

So it would be better to go parallel with the package naming convention for this folder, like: **designaleaderboard**

For detailed information:

https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

altayhunoglu (approved):
Quite fine. Thanks.

ErdemT09:
I am not sure if moving `LongestSubstringwithAtMostKDistinctCharacters` will be okay as another class is using it, that's why I've left it out.

altayhunoglu (approved):

altayhunoglu:
No problem even if you move. Another class just needs to be updated for the import part.

ErdemT09:
It has been added.

altayhunoglu:
Ready to merge then.

PR #167, opened 2021-05-05
1087. Brace Expansion

ErdemT09:
The main process of this algorithm is like this:
If we have some letters inside the braces, then the result is just of concatenation of these strings and "brace expansion" of the rest of the string.
The TreeSet is responsible for lexographically sorting these strings.

Additionally, there is also a possible solution with the list. This can also be added.
The credit for the code goes [here](https://wentao-shao.gitbook.io/leetcode/graph-search/1087.brace-expansion).

### DIFF
--- a/src/main/java/algorithms/curated170/medium/BraceExpansion.java
+++ b/src/main/java/algorithms/curated170/medium/BraceExpansion.java
@@ -1,0 +1,41 @@
+package algorithms.curated170.medium;
+
+import java.util.Arrays;
+import java.util.TreeSet;
+
+public class BraceExpansion {
+    public static String[] expand(String S) {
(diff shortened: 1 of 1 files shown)


altayhunoglu:
Hello Erdem,
 This solution is quite fine and takes 10 ms.
 As this is also a String processing question, a recursive solution also is appropriate for that problem, similar to this:
 https://github.com/spiralgo/algorithms/pull/117
 
 In this case, we would be using a StringBuilder instead of a TreeNode in order to reconstruct the string.
 
 When we are available for that hypothetic solution, we can meditate on it as well.
 
  Anyways, I approved this one. Thanks for your valuable effort and attention.

altayhunoglu (approved):

PR #168, opened 2021-05-06
1087-List solution

ErdemT09:
This new solution places the strings in a list and sorts them each time. Would it be more efficient just to sort at the end and not in every method call?

### DIFF
--- a/src/main/java/algorithms/curated170/medium/braceexpansion/BraceExpansion.java
+++ b/src/main/java/algorithms/curated170/medium/braceexpansion/BraceExpansion.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.medium;
+package algorithms.curated170.medium.braceexpansion;
 
 import java.util.Arrays;
 import java.util.TreeSet;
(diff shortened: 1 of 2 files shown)


altayhunoglu:
I think currently it is sorting at the end, and it looks fine to me.

altayhunoglu (approved):
Quite fine Erdem. Thank you.

image

ErdemT09:
I will actually remove the re-sorting of the list each time and add it. That should be faster.
```
public String[] expand(String S) {
    return expandRecursive(S, true);
  }

  private String[] expandRecursive(String S, boolean sort) {
    int n = S.length();
    if (n == 0) {
      return new String[] { "" };
    }
    if (n == 1) {
      return new String[] { S };
    }

    List<String> res = new ArrayList<>();
    if (S.charAt(0) == '{') {
      int i = 0;
      while (i < S.length()) {
        if (S.charAt(i) == '}') {
          break;
        }
        i++;
      }
      String[] leftStrs = S.substring(1, i).split(",");
      String[] rightStrs = expandRecursive(S.substring(i + 1), false);
      for (String lStr : leftStrs) {
        for (String rStr : rightStrs) {
          res.add(lStr + rStr);
        }
      }
    } else {
      String[] rightStrs = expandRecursive(S.substring(1), false);
      for (String rStr : rightStrs) {
        res.add(S.charAt(0) + rStr);
      }
    }
    if (sort) {
      Collections.sort(res);
    }
    return res.toArray(new String[0]);
  }
```

altayhunoglu:
Hello Erdem,

It is almost the same:
image

ErdemT09:
Then I am merging with the current list solution as it is simpler than the other.

PR #172, opened 2021-05-07
426. Convert Binary Search Tree to Sorted Doubly Linked List

ErdemT09:
I have tried to implement the solution that is mention in the Readme file. 
This solution utilizes a kind of "inorder" traversal.

For reference:
Depth First Traversals: 
(a) Inorder (Left, Root, Right) : 
(b) Preorder (Root, Left, Right) : 
(c) Postorder (Left, Right, Root) :

### DIFF
--- a/src/main/java/algorithms/curated170/medium/ConvertBinarySearchTreeToDoublyLinkedList.java
+++ b/src/main/java/algorithms/curated170/medium/ConvertBinarySearchTreeToDoublyLinkedList.java
@@ -1,0 +1,44 @@
+package algorithms.curated170.medium;
+
+import algorithms.datastructures.TreeNode;
+
+public class ConvertBinarySearchTreeToDoublyLinkedList {
+    private TreeNode headPointer = new TreeNode(0);
+    private TreeNode lastNode = headPointer;
+
+    public TreeNode treeToDoublyList(TreeNode root) {
+        convertTree(root);
+        lastNode.right = headPointer.right;
+        headPointer.right.left = lastNode;
+        return headPointer.right;
+    }
+
+    private void convertTree(TreeNode root) {
+        if (root == null) {
+            return;
+        }
+        convertTree(root.left);
+        lastNode.right = root;
+        root.left = lastNode;
+        lastNode = root;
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/ConvertBinarySearchTreeToDoublyLinkedList.java
```diff
@@ -0,0 +1,44 @@
...
+import algorithms.datastructures.TreeNode;
+
+public class ConvertBinarySearchTreeToDoublyLinkedList {
+    private TreeNode headPointer = new TreeNode(0);
+    private TreeNode lastNode = headPointer;
+
+    public TreeNode treeToDoublyList(TreeNode root) {
+        convertTree(root);
+        lastNode.right = headPointer.right;
+        headPointer.right.left = lastNode;
```
altayhunoglu, line 12: We need to add a null-check before using **headPointer.right** as there is a possible case in which **headPointer.right** has never been populated.

altayhunoglu: Quite a nice solution using a "sentinel node" as a dummy head.
I approve already.
To make it clear for the future audience, I draw your code as much as I understand it:

IMG-1594

altayhunoglu (approved):

src/main/java/algorithms/curated170/medium/ConvertBinarySearchTreeToDoublyLinkedList.java
ErdemT09: Does the new null check fix this problem?

altayhunoglu: Yep, it fixes like this:
(LeetCode uses Node instead of TreeNode, by the way. 
 You can also import Node using "import algorithms.datastructures.Node"
 
image

ErdemT09:
As things seem all right in the current version of the code in this PR, I am merging it.

PR #173, opened 2021-05-07
1245. Tree Diameter

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/16)
Firstly, in this question, the given tree is an undirected general tree, not necessarily a binary tree.
Let's say that some arbitrary node `n` in this tree is connected to `k` nodes.
By of this `k` nodes, there are `k` different paths to node `n` . The sum of the two longest 2 of these path will give us the diameter of the tree with respect to node `n`.
Out of all the relative node diameters, the result should be the maximum. 
We can recursively implement this, with a slight modification that we ignore the parent node that our node was called to calculate its relative diameter. 
This is the algorithm of this solution.

Note: This solution can also be implemented using a list of lists.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/TreeDiameter.java
+++ b/src/main/java/algorithms/curated170/medium/TreeDiameter.java
@@ -1,0 +1,56 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TreeDiameter {
+      private int diameter = 0;
+      private Map<Integer, List<Integer>> tree = new HashMap<>();
+
+      public int treeDiameter(int[][] input) {
+            int ret = 0;
+            for (int[] pair : input) {
+                  int a = pair[0];
+                  int b = pair[1];
+                  tree.putIfAbsent(a, new ArrayList<>());
+                  tree.putIfAbsent(b, new ArrayList<>());
+                  tree.get(a).add(b);
+                  tree.get(b).add(a);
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/TreeDiameter.java
```diff
@@ -0,0 +1,56 @@
...
+            searchDiameter(tree.keySet().iterator().next(), Integer.MIN_VALUE);
+            return diameter;
+      }
+
+      public int searchDiameter(int root, int parent) {
+            int maxArm1 = 0;
+            int maxArm2 = 0;
+            for (int k : tree.get(root)) {
+                  int depth = 0;
+                  if (k != parent) {
```
altayhunoglu, line 31: **k!=parent**:
A nice alternative to using **boolean visited[]** to prevent circle. As there is no duplicate value; we don't need an extra array.

src/main/java/algorithms/curated170/medium/TreeDiameter.java
```diff
@@ -0,0 +1,56 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TreeDiameter {
```
altayhunoglu, line 8: Quite a nice solution:
image

A similar question:
https://github.com/spiralgo/algorithms/issues/174

altayhunoglu (approved):
Approved with some informative comments.

PR #180, opened 2021-05-08
1136-Initial solution

ErdemT09:
As mentioned in the Readme file, this solution utilizes a kind of topological sorting. 
Steps are so:
1. We create two maps from the input, one for the displaying the children of the nodes, another for the parents. We also keep a note of which nodes initially have an indegree of 0.
2. Beginning from these nodes added to a queue, we add their children to the queue if the children do not have any unchecked parents left. We do this in a layered way.
3. At the end, we compare if the nodes that we have counted are any less than the total nodes, if so, that means that our graph contains some loops. We should return -1 then.
4. If not, we return the amount of layers we have calculated.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/ParallelCourses.java
+++ b/src/main/java/algorithms/curated170/medium/ParallelCourses.java
@@ -1,0 +1,78 @@
+package algorithms.curated170.medium;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Queue;
+
+public class ParallelCourses {
+    Map<Integer, HashSet<Integer>> courseChiMap;
+    Map<Integer, HashSet<Integer>> courseParMap;
+    HashSet<Integer> firstCourses = new HashSet<>();
+
+    public int solution(int[][] courses) {
+        createMap(courses);
+        return sortTopologically();
+    }
+
+    private int sortTopologically() {
+        Queue<Integer> q = new LinkedList<>();
+        q.addAll(firstCourses);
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/ParallelCourses.java
```diff
@@ -0,0 +1,78 @@
...
+
+public class ParallelCourses {
+    Map<Integer, HashSet<Integer>> courseChiMap;
+    Map<Integer, HashSet<Integer>> courseParMap;
+    HashSet<Integer> firstCourses = new HashSet<>();
+
+    public int solution(int[][] courses) {
+        createMap(courses);
+        return sortTopologically();
+    }
```
altayhunoglu, line 17: Can we refactor the code to be compatible with LeetCode?

Default code is like this:
class Solution {
    public int minimumSemesters(int n, int[][] relations) {
        
    }
}

altayhunoglu:
I updated the issue with screenshots and the default code.
https://github.com/spiralgo/algorithms/issues/18

ErdemT09:
I have added some changes to the initial solution. The alternative one uses `List<List<Integer>>`, while the other one remains with the HashMap.

altayhunoglu (approved):
Quite fine.
I approve. By the way, we can delete this part:
`   HashSet<Integer> firstCourses = new HashSet<>();
`

ErdemT09:
Should we remove the solution with the HashMap? It doesn't show anything important about this dataype and it is slower.

altayhunoglu:
Yes, we can remove.

ErdemT09:
We should also add the DFS solution at some point.

PR #181, opened 2021-05-08
510. Inorder Successor in BST II

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/19)
We have previously encountered in [this question](https://github.com/spiralgo/algorithms/issues/15), so it might be helpful to revise that before solving this.
`is`: inorder successor

The idea is simple:
If a node has a right node, then the `is` should be the leftmost node at that side of tree. 
Else, if a node has a parent node and its left node is the node, then the `is` is the node's parent. 
Else, if a node has a parent node and its right node is the node, then the `is` is the parent's parent if its parent is its corresponding left node, if its however the right node, then we must recursively climb up the tree, until we find a node that is its parent's left node.
If all these do not work, then the result is `null`, as there is no  `is`.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/InorderSuccessorInBST2.java
+++ b/src/main/java/algorithms/curated170/medium/InorderSuccessorInBST2.java
@@ -2,24 +2,24 @@ package algorithms.curated170.medium;
 
 public class InorderSuccessorInBST2 {
 
-    public Node sucNode(Node start) {
+    public Node inorderSuccessor(Node start) {
         if (start == null) {
             return null;
         }
-        if (start.right == null) {
-            while (start.parent != null) {
-                if (start.parent.left == start) {
-                    return start.parent;
-                }
-                start = start.parent;
-            }
-            return null;
+        if (start.right != null)
+            return findIS(start.right);
+
+        while (start.parent != null && start.parent.left != start) {
+
+            start = start.parent;
         }
-        return findSuc(start);
+
+        return start.parent;
+
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/InorderSuccessorInBST2.java
```diff
@@ -0,0 +1,47 @@
...
+            return null;
+        }
+        if (start.right == null) {
+            while (start.parent != null) {
+                if (start.parent.left == start) {
+                    return start.parent;
+                }
+                start = start.parent;
+            }
+            return null;
```
altayhunoglu, line 16: We can combine if and while like this:

```
  public Node inorderSuccessor(Node start) {
        if (start == null) {
            return null;
        }
      if (start.right != null) return findSuc(start.right);
       
      while (start.parent != null && start.parent.left != start) {
                
                start = start.parent;
       }
      
       return start.parent;
     
        
    }
```


image

ErdemT09:
Would an iterative solution be faster? Such as [this](https://www.junhaow.com/lc/problems/tree/bst/510_inorder-successor-in-bst-ii.html). Again, theoretically, it shouldn't be.

altayhunoglu:
> Would an iterative solution be faster? Such as [this](https://www.junhaow.com/lc/problems/tree/bst/510_inorder-successor-in-bst-ii.html). Again, theoretically, it shouldn't be.

 The iterative one is slower.
 I think recursive is a perfect match for Tree questions in general.
image

PR #182, opened 2021-05-08
369. Plus One Linked List

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/22)
Let us assume that we are given the head of list like 1-2-3. The operation on this is pretty easy. We just add 1 to the last node, the node without a next node. 
What if this number was something like 1-9? Then this operation wouldn't have worked out, as the result would have been 1-10. Hence, we should increment the last digit that is not 9 by 1, then set all the values thereafter to 0. Result is then 2-0. This algorithm also works out for 1-9-9-9, 1-0-9, 8-9-9 and such.
The last possibility is that the given list only contains 9's. Then we should have a some sort of placeholder node `p`, whose next node will be our start node. If our head node has the value 9, then the last nine node should be `p`.  In such a situation with 9-9-9-9, this node `p` will obtain the value 1. 
At the end, we return node `p` if its value is not 1,  else, we return the head.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/PlusOneLinkedList.java
+++ b/src/main/java/algorithms/curated170/medium/PlusOneLinkedList.java
@@ -4,7 +4,7 @@ import algorithms.datastructures.ListNode;
 
 public class PlusOneLinkedList {
 
-    public ListNode addOne(ListNode head)
+    public ListNode plusOne(ListNode head)
     {
         ListNode newHead = new ListNode(0);
         newHead.next = head;
--- a/src/main/java/algorithms/curated170/medium/PlusOneLinkedList.java
+++ b/src/main/java/algorithms/curated170/medium/PlusOneLinkedList.java
@@ -1,0 +1,39 @@
+package algorithms.curated170.medium;
+ 
+import algorithms.datastructures.ListNode;
+
+public class PlusOneLinkedList {
+
+    public ListNode addOne(ListNode head)
+    {
+        ListNode newHead = new ListNode(0);
+        newHead.next = head;
+        ListNode lastNonNine = head.val == 9 ? newHead : head;
+        ListNode curr = head;
+        while(curr != null)
+        {
+            if(curr.val != 9)
+            {
+                lastNonNine = curr;
+            }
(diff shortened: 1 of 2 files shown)


ErdemT09:
In these questions with nodes, graphs and such, I am strongly seeing some aspects of Turing Machines. I am not really well-versed in this topic but that are my observations.

src/main/java/algorithms/curated170/medium/PlusOneLinkedList.java
```diff
@@ -0,0 +1,39 @@
+package algorithms.curated170.medium;
+ 
+import algorithms.datastructures.ListNode;
+
+public class PlusOneLinkedList {
+
+    public ListNode addOne(ListNode head)
```
altayhunoglu, line 7: Can we change the method name to "plusOne" in order to be compatible with LeetCode?

altayhunoglu: Quite a smart solution using sentinel node as a dummy head.
image

altayhunoglu (approved):

ErdemT09:
Another sentinel node question. This actually goes together with [this](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/medium/ConvertBinarySearchTreeToDoublyLinkedList.java). Perhaps as a preparation question for it.

altayhunoglu:
> Why can't I merge this PR?

There is a conflict with the code in master.
Can you try to pull master and merge it with the master.
Then push again, and let us see what will happen.

ErdemT09:
I guess you have used Git commands to merge it. I did not understand what the conflict here was, but I guess the solution you have mentioned would also have worked.

altayhunoglu:
> Note: In these questions with nodes, graphs and such, I am strongly seeing some aspects of Turing Machines. I am not really well-versed in this topic but that are my observations.

Yep. We can assume that the Turing machine is the basis of all modern computers.
Perhaps it is possible to think of a Turing machine as a binary decision tree that is constructed as a boolean array.

PR #183, opened 2021-05-08
281. Zigzag Iterator

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/23)
As stated in the Readme, this solution uses a queue of pointers, if we consider the iterators to be pointers. We add the iterators to the given lists to a queue. We poll one iterator first, retrieve its value, put it back to the queue if its not empty. we can also handle arbitrarily many given lists with this algorithm. We do not need to check if the queue is empty in the `next()` method because it is guaranteed that it won't be empty when it is called, similarly to how I have implemented an example in the main method: 

> By calling  next repeatedly until hasNext returns false


altayhunoglu:
Quite satisfying and nothing to add.

altayhunoglu (approved):
Quite satisfying and nothing to add.
image

ErdemT09:
Is the solution [here](https://medium.com/@rebeccahezhang/leetcode-281-zigzag-iterator-a5ee60dd20ea) faster? It shouldn't be, I think.

altayhunoglu:
> Is the solution [here](https://medium.com/@rebeccahezhang/leetcode-281-zigzag-iterator-a5ee60dd20ea) faster? It shouldn't be, I think.

It is slower and more complex I think:
image

ErdemT09:
Perhaps Java's built in `Iterator` class is slow. A pointer solution as [this](https://tenderleo.gitbooks.io/leetcode-solutions-/content/GoogleMedium/281.html) can perhaps be faster, but it won't hold for the follow-up question in LeetCode.

altayhunoglu:
> Perhaps Java's built in `Iterator` class is slow. A pointer solution as [this](https://tenderleo.gitbooks.io/leetcode-solutions-/content/GoogleMedium/281.html) can perhaps be faster, but it won't hold for the follow-up question in LeetCode.
> We should also test this on LeetCode.
It is the result:

image

altayhunoglu:
> Perhaps Java's built in `Iterator` class is slow. A pointer solution as [this](https://tenderleo.gitbooks.io/leetcode-solutions-/content/GoogleMedium/281.html) can perhaps be faster, but it won't hold for the follow-up question in LeetCode.
> We should also test this on LeetCode.

I think the latest code you mentioned is quite clean and maintainable for the follow-up.
We can first refactor the current code to be maintainable just removing the need for the "indices" array.

REFACTOR:
```
 public class ZigzagIterator {               
     List<List<Integer>> lists = new ArrayList<>();                                 
      int p = 0; // which list                 
      int q = 0; // current index in current list;                                   
                                 
      public ZigzagIterator(List<Integer> v1, List<Integer> v2) {                    
       lists.add(v1);   
       lists.add(v2);                  
                    
      }                                                                              

      public int next() {                                                            
          int val = lists.get(p).get(q);                                             
          incrementPointers();                                                            
          return val;                                                                
      }                                                                              

      public boolean hasNext() {                                                     
          if(lists.get(p).size() > q) return true;                          
          else{                                                                      
             
              incrementPointers();                                          
              return lists.get(p).size() > q;                                        
          }                                                                          
      }    
       void incrementPointers() {
              p = 1-p;   
              if(p==0){
                   q++;
              } 
       }
  }  
```                
It takes:
image

After that, we can alter the code to be maintainable using "p" as a pointer that can increase till the end of the "list"  (that holds vectors), rather than just as a binary flag switching between "p" and "1-p".

Nevertheless, in this turn, we need to count errors. As each row has the right to get an error just once, we can use a set for marking the row of which's size limit has already been exceeded.

MAKE IT MAINTAINABLE: (and it also takes 1-2 ms.)
```
class ZigzagIterator {

    List<List<Integer>> lists = new ArrayList<>();
    int p = 0; // which list                 
    int q = 0; // current index in current list;                                   
    int rows = 0;

    public ZigzagIterator(List<Integer> v1, List<Integer> v2) {
        lists.add(v1);
        lists.add(v2);
        rows = lists.size();
    }

    public int next() {
        int val = -1;

        if (hasNext()) {
            val = lists.get(p).get(q);
            incrementPointers();
        }

        return val;
    }

    public boolean hasNext() {

        if (lists.get(p).size() > q) {
            return true;
        }
        countError.add(p);
        if (incrementPointers()) {
            return hasNext();
        }
        return false;

    }
   HashSet countError = new HashSet();

    boolean incrementPointers() {
        if (countError.size() < rows) {

            if (++p == rows) {
                p = 0;
                q++;
            }
            return true;
        }
        return false;
    }
}
```


To test the maintainable code, we can have a main method like this.

```
        List<Integer> v1 = Arrays.asList(1, 6, 11),
                v2 = Arrays.asList(2, 7, 12),
                v3 = Arrays.asList(3, 8),
                v4 = Arrays.asList(4, 9, 13, 15),
                v5 = Arrays.asList(5, 10, 14, 16);
        ZigzagIterator z = new ZigzagIterator(v1, v2);
        z.append(v3);
        z.append(v4);
        z.append(v5);

        int val;
        do {
            val = z.next();
            System.out.println(val);
        } while (val > -1);
    }
```

Adding an extra append method into ZigzagIterator to test : 
```
  public void append(List<Integer> extra) {
        lists.add(extra);
        rows = lists.size();
    }
```

ErdemT09:
Should we keep the solution with the Iterator or remove it?

altayhunoglu:
> Should we keep the solution with the Iterator or remove it?

 I think we should keep it as it is more straightforward and it is why appropriate for a usual interview.

PR #184, opened 2021-05-09
1151. Minimum Swaps to Group All 1's Together

ErdemT09:
In the given array, all 1's must be grouped together. This is logically equivalent to saying that all 1's should be located in a given window in the array, length of which is the number of 1's in the array.
We count that there are 5 1's in the array. 
For example there is this window starting at some index: 1, 1, 0, 0, 1
This means that we need to swap out the 2 0's in order to have only 1's in this window, and hence all the 1's together.
With this procedure, we check all the windows in the array and return the least amount of 0's that is counted in any of these windows.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/MinimumSwaps.java
+++ b/src/main/java/algorithms/curated170/medium/MinimumSwaps.java
@@ -14,7 +14,7 @@ public class MinimumSwaps {
                   return 0;
             }
             int minChanges = windowSize;
-            for(int i = 0; i<data.length-windowSize; i++)
+            for(int i = 0; i<=data.length-windowSize; i++)
             {
                   minChanges = Math.min(minChanges, countZeros(i));
             }
--- a/src/main/java/algorithms/curated170/medium/MinimumSwaps.java
+++ b/src/main/java/algorithms/curated170/medium/MinimumSwaps.java
@@ -1,43 +1,27 @@
 package algorithms.curated170.medium;
 
 public class MinimumSwaps {
-      private int[] data;
-      private int windowSize = 0;
 
-      public int minimumSwaps(int[] data) {
-            this.data = data;
+      public int minSwaps(int[] data) {
+            int windowSize = 0;
+
             for (int i : data) {
-                  if (i == 1) {
-                        windowSize++;
-                  }
-            }
-            if (windowSize == 0 || windowSize == 1) {
-                  return 0;
+                  windowSize += i;
             }
 
             int minChanges = 0;
             for (int i = 0; i < windowSize; i++) {
-                  System.out.println(i + " " + data[i]);
-                  if (data[i] == 0) {
-                        minChanges++;
-                  }
+
+                  minChanges += (1 - data[i]);
+
             }
 
-            System.out.println(windowSize + " " + minChanges);
             int prev = minChanges;
-            for (int l = 1, r = windowSize; l <= data.length - windowSize; l++, r++) {
-                  int a = data[l-1], b = data[r];
-                  System.out.println("index: " + l + " prev: " + prev + " a: " + a + " b: " + b + " minChanges: " + minChanges);
-                  if (a == b) {
-                        continue;
-                  }
-                  
-                  if (a > b) {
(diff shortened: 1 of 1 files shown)


altayhunoglu:
**SPOILER ALERT:**
  
  I will review the code when I am available.
  By the way, I will give some spoilers for another approach:
  
 
 - Traverse whole the array, shifting the window from left to right in order to detect a subarray that contains the maximum number of ones (maxOnesCount).
 - Then let us substruct maxOnesCount from the size of the window, and here is the result.
 
  We can simulate shifting the window over the data array, using a left and a right pointer. 
  (It would be much more similar to a caterpillar's movement in which we try to preserve the size of the window. )
  We will increase only the right pointer until the distance between the right and the left (right-left) reaches the windowSize. 
  
  Then we will increase the left pointer to preserve windowSize. 
  While increasing the right pointer, we can also count the ones in the current window like this:
   currentOnes+= data[right++];
   
   We can also calculate windowSize easier like this: (though we should compare efficiency)
   int windowSize = Arrays.stream(data).sum();

ErdemT09:
That's sort of a dynamic peogramming approach. I shall implement it then.

altayhunoglu:
> That's sort of a dynamic programming approach. I shall implement it then.
> Edit: Added this approach, although a bit different.

 When I comprehend your approach completely, I see that it is a brilliant idea to create a window initially.
 In this way, we get rid of extra time to ask if **right>windowSize**.
 
Though, the code needed some refactoring and I will try to fulfill it here, step by step:
 
 **1- Remove unnecessary prints and if-statements. As there isn't another method in the class, we do not need global variables as well.**
 
```
 public class Solution {
      public int minSwaps(int[] data) {
         private int windowSize = 0;
            for (int i : data) {
                        windowSize += i;
            }
            int minChanges = 0;
            for (int i = 0; i < windowSize; i++) {
                 minChanges += (1 - data[i]);
             }

            int prev = minChanges;
            
          for (int left = 1, right = windowSize; left <= data.length - windowSize; left++, right++) {
                  int a = data[left-1], b = data[right];
                  if (a > b) {
                        prev++;
                  } else if (a < b) {
                        prev--;
                  }
                   minChanges = Math.min(minChanges, prev);
            }
            return minChanges;
      }
} 
```
 
2- Rethink about "a>b" and "b>a":
  
   As both a and b are binary, then:
```
      if (a > b) {
                        prev++;
                  } else if (a < b) {
                        prev--;
                  }
```
 equals to:
   
```
     if (a == 1 && b == 0) {
                        prev++;
                  } else if (a == 0 && b == 1) {
                        prev--;
                  }
```

  It also means; we increment `prev` by **a** and we decrement by **b**.
  So we can simply put that:
 
```
         prev += data[left-1];
         prev -= data[right];
 
```

3- Rethink about **left** and **right**.

 As we try to preserve windowSize, `left` will always mean `right-windowSize` in this case.
 So, indeed we can handle this with just one pointer called "i"
 Therefore we can alter this:
 ```

          prev += data[i-windowSize-1];
           prev -= data[i];
```
  
   
   As we make "left" obsolete, then we can rethink about for-loop like this: 

```
      for (int i= windowSize; i < data.length; i++) { 
                      prev += data[i-windowSize] - data[i];
                       
                        minChanges = Math.min(minChanges, prev);
            }
```

4- The latest refactored version would be like that:

```

public class Solution {
   public int minSwaps(int[] data) {
           int windowSize = 0;

            for (int i : data) {
                        windowSize += i;
            }
 
            int minChanges = 0;
            for (int i = 0; i < windowSize; i++) {
 
                   minChanges +=   (1-data[i]);
                  
            }

            int prev = minChanges;
            
          for (int i= windowSize; i < data.length; i++) { 
                 
                        prev += data[i-windowSize] - data[i];
                       
                        minChanges = Math.min(minChanges, prev);
            }
            return  minChanges;
      }
 } 

```

And, in the end, it takes remarkably lesser than the others:

image


Thank you for the brilliant idea and your valuable collaboration, Erdem.

altayhunoglu (approved):

PR #188, opened 2021-05-10
1166. Design File System

ErdemT09:
In this solution, we basically separate the folder names and then check if folders without the last one are in the HashMap. If so, we add the given value with all folder names listed in the HashMap. 
For the implementation of the `get( )` method, we just try to get the corresponding value in a similar way from the HashMap.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/DesignFileSystem.java
+++ b/src/main/java/algorithms/curated170/medium/DesignFileSystem.java
@@ -1,0 +1,43 @@
+package algorithms.curated170.medium;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DesignFileSystem {
+ 
+
+    Map<List<String>, Integer> files;
+    public DesignFileSystem()
+    {
+        files = new HashMap<>();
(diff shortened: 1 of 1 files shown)


ErdemT09:
http://code.minggr.net/leetcode-1166/
This node approach is also noteworthy. We should also implement this in Java.

ErdemT09:
https://jojoyang.site/2020/10/30/1166-Design-File-System/
How does this work even though it does no have any standard control checks in its method?

altayhunoglu:
> > ** if parent not in self.d or path in self.d: return False ** It is the control I think.
> 
> We also do the same in Java. Is controlling if the path is already created also necessary? Besides we have these two controls:
> 
> ```
>  if (path.length() < 2 || path.charAt(0) != '/') {
>             return false;
>         }
> ```
> 
> How does the Python solution work without these?

It is my fault that I thought "/" cannot be a valid path. But it is. Therefore we do not need any control for path.charAt(0) != '/' at the beginning.

rfind gives what is before the latest "/" and it means the parent.
So, it uses the initial value of self.d both for the standard controls and as a holder.

altayhunoglu (approved):

PR #189, opened 2021-05-10
249-A small change suggestion

ErdemT09:
My suggestion is that, instead of comparing the value of each char from the previous char, we can just compare them to the first char to get the same hash for different "shifted" strings .
Consider: "acfz" => This would correspond to "2,5,25"; so will be the string "zbey". As the difference from the first char must be the same in both strings if they are "shifted".

Additionally, I have also done some auto-formatting. That's why there seems to be these many changes.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/GroupShiftedStrings.java
+++ b/src/main/java/algorithms/curated170/medium/GroupShiftedStrings.java
@@ -5,31 +5,31 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
- 
 public class GroupShiftedStrings {
- 
-  public List<List<String>> groupStrings(String[] strings) {
-    Map<String, List<String>> map = new HashMap<>();
 
-    for(String s : strings) {
-        String key = getKey(s);
-        List<String> list = map.getOrDefault(key, new ArrayList<>());
-        list.add(s);
-        map.put(key, list);
-    }
-    return new ArrayList<>(map.values());
-}
+    public List<List<String>> groupStrings(String[] strings) {
+        Map<String, List<String>> map = new HashMap<>();
 
-private String getKey(String s) {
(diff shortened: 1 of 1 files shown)


altayhunoglu:
Although both solutions take the same time (9ms), using the first char as a standpoint is a smart idea.
 
 Another approach is, to calculate the difference for the first char and apply it to all: (It takes 1ms)
 
```
   public String getKey(String str) {
        int diff = str.charAt(0) - 'a';
        char[] chars = str.toCharArray();
        for (int i = 0; i < chars.length; i++) {
            chars[i] = chars[i] - diff < 'a' ? (char)(chars[i] - diff + 26) : (char)(chars[i] - diff);
        }
        return String.valueOf(chars);
    }
```
In this picture, diff = 25.
Then we apply the same subtraction to all chars and estimate their first state (before any shifting process)

IMG-1598

ErdemT09:
I think that we can actually get rid of the ternary operator by inserting the modulus operator:
 public String getKey(String str) {
        int diff = str.charAt(0) - 'a';
        char[] chars = str.toCharArray();
        for (int i = 0; i < chars.length; i++) {
            chars[i] = (char)((chars[i] - diff + 26)%26);
        }
        return String.valueOf(chars);
    }

Is this better?

altayhunoglu:
> public String getKey(String str) {
>         int diff = str.charAt(0) - 'a';
>         char[] chars = str.toCharArray();
>         for (int i = 0; i < chars.length; i++) {
>             chars[i] = (char)((chars[i] - diff + 26)%26);
>         }
>         return String.valueOf(chars);
>     }

Yep. Great idea Erdem.

We can take this.

image

ErdemT09:
Applying it then.

altayhunoglu (approved):
Great refactoring. Thanks.

PR #190, opened 2021-05-10
1136. Parallel Courses DFS

ErdemT09:
This question is also answerable through the usage of DFS:
Explanation:
1. We create a Map, showing which child values a parent value has (`HashMap<Integer, List<Integer>> graph`)
2.  We then apply this DFS algorithm on the courses:
We search down through each node until we reach the last obtainable node. Last obtainable node returns a distance value of 1, because its `max` value will stay 0 and it will increment this by one.
After that, last nodes neighbors increment their max value in a similar way to 1, because they do not have any other node that could have a higher return value than 1. 
This recursive method returns itself up until the node, whereat we started the DFS. If we encounter any other nodes, whose depth we have calculated in this DFS, we would know their depth as we store them in the array `depth[ ]`.
We know if there are any loops present if we encounter the same node while searching. We can know this by uing the boolean array `visited[ ]`. 
If we encounter a cycle, we set the boolean `hasCycle` to true and return the wanted result of -1.
Else, we return the maximum depth that we encounter.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/parallelcourses/ParallelCoursesBFS.java
+++ b/src/main/java/algorithms/curated170/medium/parallelcourses/ParallelCoursesBFS.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.medium;
+package algorithms.curated170.medium.parallelcourses;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


altayhunoglu (approved):
Great solution.
Detecting cycle and calculating graph diameter in the same DFS method is very nice indeed.

Apparently, this solution is relatively close to your solution for Tree Diameter:
https://github.com/spiralgo/algorithms/pull/173

It is because **the length of the longest path** is the diameter of a tree (which is a kind of graph) and it also gives us the process partitions (moves, steps, or semesters in that context).

For our Turkish Black Tea graph, the longest path is 5.
So we need at least 5 interdependent steps to make a tea. 
We can do 3 simultaneously with any node before 4 as it is not dependent on any other steps. 
(Turkish tea making graph from: https://www.youtube.com/watch?v=99oTqO51Jh0)

IMG-1599

PR #195, opened 2021-05-11
1057. Campus Bikes

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/32)
There are 2 solutions two this question. The first one is the "bucket sort" method as mentioned in the Readme file. The other one is a priority queue.

First solution-Bucket Sort:
The maximum possible dimension of the grid are 1000 X 1000, which means that the maximum distance is 2000.
This means that we can store all possible worker-bike pairs with respect to their distances in an array of length 20.

After this procedure check each List of pairs  (bucket) in this array (buckets). we traverse this array naturally from 0 to its end.
In each bucket, if we have not encountered this bike and the worker, that means that these bike and the worker are their pair. We store the bike's index in an array.
The setting by indexing is also taken care of, as we built the buckets iteratively from the workers and the bikes.
We return the stored bike indices at the end.

Second solution-Priority Queue:
Here, we use a priority queue that stores a triplet* of the numbers ( distance between bike and worker, worker id, bike id). Similarly, we iterate through each bike and worker and add the relative triplet to the priority queue.
After creating the priority queue, the items we poll from it are compared first by their distances. If same, by the worker ids, if same, by the bike ids. We poll each item from the priority queue until we have assigned all the workers a bike. In order to avoid multiple worker-bike pairs interfering, we keep two checkers for this as we had done in the previous question.

*https://en.wikipedia.org/wiki/Tuple Learning about the naming here is helpful.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
+++ b/src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
@@ -1,7 +1,6 @@
 package algorithms.curated170.medium;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 public class CampusBikesBucketSort {
@@ -10,7 +9,7 @@ public class CampusBikesBucketSort {
       List<int[]>[] buckets = new List[2001];
 
       public int[] assignBikes(int[][] workers, int[][] bikes) {
-            
+
             w = workers.length;
 
             assignToBuckets(workers, bikes);
@@ -20,19 +19,18 @@ public class CampusBikesBucketSort {
 
       private void assignToBuckets(int[][] workers, int[][] bikes) {
             for (int i = 0; i < w; i++) {
-                  for (int j = 0; j < w; j++) {
-                        int dis = Math.abs(bikes[j][0] - workers[i][0]) + Math.abs(bikes[j][1] - workers[i][1]);
-                        if (buckets[dis] == null) {
-                              buckets[dis] = new ArrayList<>();
+                  for (int j = 0; j < bikes.length; j++) {
+                        int dist = dist(workers[i], bikes[j]);
+                        if (buckets[dist] == null) {
+                              buckets[dist] = new ArrayList<>();
                         }
-                        buckets[dis].add(new int[] { i, j });
+                        buckets[dist].add(new int[] { i, j });
                   }
             }
       }
 
-      private int[] assignFromBuckets()
-      {
-            int[] res = new int[w];
+      private int[] assignFromBuckets() {
+            int[] wo = new int[w];
             boolean[] bikeAssigned = new boolean[w];
             boolean[] workerAssigned = new boolean[w];
             short assigned = 0;
@@ -40,16 +38,25 @@ public class CampusBikesBucketSort {
             for (int i = 0; i < buckets.length && assigned < w; i++) {
                   if (buckets[i] != null) {
                         for (int[] pair : buckets[i]) {
-                              if (!workerAssigned[pair[0]] && !bikeAssigned[pair[1]]) {
-                                    res[pair[0]] = pair[1];
-                                    bikeAssigned[pair[1]] = true;
-                                    workerAssigned[pair[0]] = true;
+
+                              int w = pair[0];
+                              int b = pair[1];
+
+                              if (!workerAssigned[w] && !bikeAssigned[b]) {
+                                    wo[w] = pair[b];
+
+                                    bikeAssigned[b] = true;
+                                    workerAssigned[w] = true;
                                     assigned++;
                               }
                         }
                   }
             }
-            return res;
+            return wo;
+      }
+
+      private int dist(int[] w, int[] b) {
+            return Math.abs(w[0] - b[0]) + Math.abs(w[1] - b[1]);
       }
 
       public static void main(String[] args) {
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+
+public class CampusBikesBucketSort {
+
+      int w;
+      List<int[]>[] buckets = new List[2001];
+
+      public int[] assignBikes(int[][] workers, int[][] bikes) {
+            
+            w = workers.length;
```
altayhunoglu, line 14: As there may be more bikes then workers, we need to have variable for bikes' size also.
```
            workerSize = workers.length;
            bikeSize = bikes.length;
```

src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
```diff
@@ -0,0 +1,58 @@
...
+
+            assignToBuckets(workers, bikes);
+
+            return assignFromBuckets();
+      }
+
+      private void assignToBuckets(int[][] workers, int[][] bikes) {
+            for (int i = 0; i < w; i++) {
+                  for (int j = 0; j < w; j++) {
+                        int dis = Math.abs(bikes[j][0] - workers[i][0]) + Math.abs(bikes[j][1] - workers[i][1]);
```
altayhunoglu, line 24: In general, your solution is remarkably modular and very nicely organized.

Therefore we can also modularize the distance function as it has no direct dependence on the assignToBucketsFunction in which it will be injected.

```
  private int dist(int[] w, int[] b) {
        return Math.abs(w[0] - b[0]) + Math.abs(w[1] - b[1]);
    }
```

src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
```diff
@@ -0,0 +1,58 @@
...
+                  }
+            }
+      }
+
+      private int[] assignFromBuckets()
+      {
+            int[] res = new int[w];
+            boolean[] bikeAssigned = new boolean[w];
+            boolean[] workerAssigned = new boolean[w];
+            short assigned = 0;
```
altayhunoglu, line 38: I could not grasp why the **assigned** variable matters. We can simply remove it.

src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
```diff
@@ -0,0 +1,58 @@
...
+
+            for (int i = 0; i < buckets.length && assigned < w; i++) {
+                  if (buckets[i] != null) {
+                        for (int[] pair : buckets[i]) {
+                              if (!workerAssigned[pair[0]] && !bikeAssigned[pair[1]]) {
```
altayhunoglu, line 43: Instead of reading from the array in each iteration more than once, we can assign the value to a variable once:
```
    int w = pair[0];
    int b = pair[1];
```

src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesPriorityQueue.java
```diff
@@ -0,0 +1,61 @@
+package algorithms.curated170.medium;
+
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.PriorityQueue;
```
altayhunoglu, line 5: We can apply the same refactoring with the Buckert Sort solution.

src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
ErdemT09: workerAssigned is an array, not a list. That's why I needed a variable like that.

altayhunoglu: Yeah, you are right.

src/main/java/algorithms/curated170/medium/campusbikes/CampusBikesBucketSort.java
```diff
@@ -0,0 +1,93 @@
...
+                        
+                        buckets[dist].add(new int[] { i, j });
+                  }
+            }
+      }
+
+      private int[] assignFromBuckets() {
+            int[] wo = new int[w];
+            boolean[] bikeAssigned = new boolean[w];
+            boolean[] workerAssigned = new boolean[b];
```
altayhunoglu, line 38: If we fix these:
```
            boolean[] bikeAssigned = new boolean[b];
            boolean[] workerAssigned = new boolean[w];
```
image

ErdemT09: I have fixed it, actually at the same moment without seeing this comment.

ErdemT09: How about the Priority Queue solution?

altayhunoglu: It is a little bit slow.
But, maybe it is normal for PQ.
image

ErdemT09:
This question is a good example of modularity helping in managing code. If it hasn't been for the fact that we have separated the code into multiple methods, it would have been harder for us to debug it.
Besides, making such attention mistakes is something that we should not do again in the future.

altayhunoglu:
Absolutely.

altayhunoglu (approved):
Modular and clear code.

PR #196, opened 2021-05-11
1136-Modularity and optimizations

ErdemT09:
In both solutions, I removed unnecessary nested procedures to individual methods, making the code more modular.
I have also replaced the HashMap in the DFS solution with an ArrayList as we had done with the BFS solution.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/parallelcourses/ParallelCoursesBFS.java
+++ b/src/main/java/algorithms/curated170/medium/parallelcourses/ParallelCoursesBFS.java
@@ -1,20 +1,16 @@
 package algorithms.curated170.medium.parallelcourses;
 
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Queue;
 
-public class ParallelCourses {
+public class ParallelCoursesBFS {
     List<List<Integer>> courseChiMap;
     int[] inDegreeMap;
     int numOfCourses;
 
     public int minimumSemesters(int n, int[][] courses) {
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/medium/parallelcourses/ParallelCoursesBFS.java
```diff
@@ -50,7 +46,7 @@ private int sortTopologically() {
     }
 
     private void createMap(int[][] courses) {
-
+        inDegreeMap = new int[numOfCourses + 1];
         courseChiMap = new LinkedList<>();
```
altayhunoglu, line 50: We should convert this to:
     courseChiMap = new ArrayList<>();
 
 Otherwise, it will take 85ms instead of 8ms.

src/main/java/algorithms/curated170/medium/parallelcourses/ParallelCoursesDFS.java
```diff
@@ -1,36 +1,49 @@
...
+
+        createMap(n, relations);
+
+        searchNodes(n);
+
+        return hasCycle ? -1 : maxPathLength;
+    }
+
+    private void createMap(int n, int[][] relations) {
+        courseChiMap = new LinkedList<>();
```
altayhunoglu, line 26: We should also change this to:
 courseChiMap = new ArrayList<>();
 
 Otherwise it will take 80ms instead of 5ms.

ErdemT09: Added.

altayhunoglu (approved):
Thanks for refactoring. Quite clear.

PR #197, opened 2021-05-12
694. Number of Distinct Islands

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/30)
To understand this question, it is important to have understood the question  [200. Number of Islands](https://github.com/spiralgo/algorithms/issues/97). 
Each time we encounter an island, we encounter it from its up-left corner, so a calculation made according to this would hold valid for all cases. Then similar islands can be denoted using their relation to this original node. Consider the examples in the image:
image
Here, we can see that other islands are sub-islands with respect to the original node, and these sub-islands also have their own sub-islands. With this thinking, we can encode this naming through a DFS algorithms as well as a BFS one. I chose DFS because it was similar to the previous question and it seemed easier to wrap one's head around.
Note that a single piece of island here is actually described here as empty string "".
So, an island like 1 1 is actually right().
We put these island names in a Set and return this set's size.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
+++ b/src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
@@ -5,25 +5,22 @@ import java.util.HashSet;
 import java.util.Set;
 
 public class NumberOfDistinctIslands {
-    
+
     Set<String> islands;
     int[][] grid;
     int xDim;
     int yDim;
-    public int numOfIslands(int[][] grid)
-    {
+
+    public int numDistinctIslands(int[][] grid) {
         this.grid = grid;
         islands = new HashSet<>();
         xDim = grid.length;
         yDim = grid[0].length;
 
-        for(int i = 0; i<xDim; i++)
-        {
-            for(int j = 0; j<yDim; j++)
-            {
+        for (int i = 0; i < xDim; i++) {
+            for (int j = 0; j < yDim; j++) {
                 String k = DFS(i, j);
-                if(k!="#")
-                {
+                if (k != "#") {
                     islands.add(k);
                 }
             }
@@ -31,58 +28,43 @@ public class NumberOfDistinctIslands {
 
         return islands.size();
     }
-    
-    private String DFS(int i, int j)
-    {
-        String islandCode = "";
-        if(grid[i][j] == 0)
-        {
+
+    private String DFS(int i, int j) {
+        StringBuilder islandCode = new StringBuilder("");
+        if (grid[i][j] == 0) {
             return "#";
         }
         grid[i][j] = 0;
-        if(i<xDim-1&&grid[i+1][j] == 1)
-        {
-            islandCode += "d(";
-            islandCode += DFS(i+1, j);
-            islandCode += ")";
+        if (i < xDim - 1 && grid[i + 1][j] == 1) {
+            islandCode.append("d(");
+            islandCode.append(DFS(i + 1, j));
+            islandCode.append(")");
         }
-        if(j<yDim-1&&grid[i][j+1] == 1)
-        {
-            islandCode += "r(";
-            islandCode += DFS(i, j+1);
-            islandCode += ")";
+        if (j < yDim - 1 && grid[i][j + 1] == 1) {
+            islandCode.append("r(");
+            islandCode.append(DFS(i, j + 1));
+            islandCode.append(")");
         }
-        if(i>0&&grid[i-1][j] == 1)
-        {
-            islandCode += "u(";
-            islandCode += DFS(i-1, j);
-            islandCode += ")";
+        if (i > 0 && grid[i - 1][j] == 1) {
+            islandCode.append("u(");
+            islandCode.append(DFS(i - 1, j));
+            islandCode.append(")");
         }
-        if(j>0&&grid[i][j-1] == 1)
-        {
-            islandCode += "l(";
-            islandCode += DFS(i, j-1);
-            islandCode += ")";
+        if (j > 0 && grid[i][j - 1] == 1) {
+            islandCode.append("l(");
+            islandCode.append(DFS(i, j - 1));
+            islandCode.append(")");
         }
-        return islandCode;
+        return islandCode.toString();
     }
 
-
     public static void main(String[] args) {
-        int[][] grid =  new int[][]{
-            {1, 1, 1},
-            {0, 1, 0},
-            {0, 1, 0},
-            {0, 0, 1},
-            {0, 0, 1},
-            {1, 1, 0},
-            {1, 0, 1},
-            {0, 1, 1},
-            {1, 0, 0}
-        };
-        
+        int[][] grid = new int[][] { { 1, 1, 1 }, { 0, 1, 0 }, { 0, 1, 0 }, { 0, 0, 1 }, { 0, 0, 1 }, { 1, 1, 0 },
+                { 1, 0, 1 }, { 0, 1, 1 }, { 1, 0, 0 } };
+
         System.out.println(Arrays.deepToString(grid));
         var solution = new NumberOfDistinctIslands();
-        solution.numOfIslands(grid);
+        solution.numDistinctIslands(grid);
+        System.out.println(solution.islands);
     }
 }
--- a/src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
+++ b/src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
@@ -7,64 +7,87 @@ import java.util.Set;
 public class NumberOfDistinctIslands {
 
     Set<String> islands;
-    int[][] grid;
+    int[][] grid; 
     int xDim;
     int yDim;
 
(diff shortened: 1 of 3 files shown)


ErdemT09:
Another option is to encode the island names by the distance of the island piece from the parent node:
The is, for example:
0 0 1 0
0 1 1 0
the list of pairs: (0,0), (0, -1), (-1, -1) .

src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+public class NumberOfDistinctIslands {
+    
+    Set<String> islands;
+    int[][] grid;
+    int xDim;
+    int yDim;
+    public int numOfIslands(int[][] grid)
```
altayhunoglu, line 13: Can we alter the method name as numDistinctIslands to be more compatible with LeetCode?

altayhunoglu:
Updated issue with images and default code:
https://github.com/spiralgo/algorithms/issues/30

src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
```diff
@@ -0,0 +1,88 @@
...
+    
+    private String DFS(int i, int j)
+    {
+        String islandCode = "";
+        if(grid[i][j] == 0)
+        {
+            return "#";
+        }
+        grid[i][j] = 0;
+        if(i<xDim-1&&grid[i+1][j] == 1)
```
altayhunoglu, line 43: We should reduce if statements into a single base case as we have done in:
 https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/medium/NumberOfIslands.java

src/main/java/algorithms/curated170/medium/NumberOfDistinctIslands.java
```diff
@@ -0,0 +1,88 @@
...
+            islandCode += ")";
+        }
+        if(j>0&&grid[i][j-1] == 1)
+        {
+            islandCode += "l(";
+            islandCode += DFS(i, j-1);
+            islandCode += ")";
+        }
+        return islandCode;
+    }
```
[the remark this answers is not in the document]
ErdemT09: They are indeed faster than reconstructing a new String object every single time, you're right.

altayhunoglu:
In LeetCode solution, as far as I see, there is no "(".
 It means we can do it too.
- It is why I tried to reduce the base case first, then I removed the parentheses. I also works like that.
 
```
 public class Solution {

       Set<String> islands;
    int[][] grid;
    int xDim;
    int yDim;

    public int numDistinctIslands(int[][] grid) {
        this.grid = grid;
        islands = new HashSet<>();
        xDim = grid.length;
        yDim = grid[0].length;

        for (int i = 0; i < xDim; i++) {
            for (int j = 0; j < yDim; j++) {
                String k = DFS(i, j);
                if (k != "#") {
                    islands.add(k);
                }
            }
        }

        return islands.size();
    }

    private String DFS(int i, int j) {
        StringBuilder islandCode = new StringBuilder("");
      if (i < 0 || j < 0 || i >= grid.length || j >= grid[0].length)   return "#"; 
       if (grid[i][j] == 0) {
            return "#";
        }
        grid[i][j] = 0;
         
            islandCode.append("d");
            islandCode.append(DFS(i + 1, j));
          
            islandCode.append("r");
            islandCode.append(DFS(i, j + 1));
      
            islandCode.append("u");
            islandCode.append(DFS(i - 1, j));
          
            islandCode.append("l");
            islandCode.append(DFS(i, j - 1));
           
     
        return islandCode.toString();
    }
 
}

```

- If we want to get rid of these appends:
```
public class Solution {

    Set<String> islands;
    int[][] grid; 
     
    int xDim;
    int yDim;
    public int numDistinctIslands(int[][] grid)
    {
        this.grid = grid; 
        islands = new HashSet<>();
        xDim = grid.length;
        yDim = grid[0].length;

        for(int i = 0; i<xDim; i++)
        {
            for(int j = 0; j<yDim; j++)
            {
               islandCode = new StringBuilder();
                 DFS(i, j, "#");
                  if (islandCode.length() == 0) {
                    continue;
                }
                islands.add(islandCode.toString());
            }
        }

        return islands.size();
    }
    StringBuilder islandCode = new StringBuilder();
    private void DFS(int i, int j, String dir)
    {
      if (i < 0 || j < 0 || i >= grid.length || j >= grid[0].length || grid[i][j] == 0) return;    
            
          grid[i][j] = 0;
       
           islandCode.append(dir);
            DFS(i+1, j, "d");
            DFS(i-1, j, "u");
            DFS(i, j+1, "r");
            DFS(i, j-1, "l");
            islandCode.append("#");
    
    }

 
}


```

ErdemT09:
For this grid:
```
int[][] grid = new int[][] { 
        { 1, 1, 1 },
        { 0, 1, 0 }, 
        { 0, 1, 0 },
        { 0, 0, 1 }, 
        { 0, 0, 1 },
        { 1, 1, 0 },
        { 1, 0, 1 },
        { 0, 1, 1 }, 
        { 1, 0, 0 } };
```
The islands are `[#d##, ##, #rdd##r###, #dl###, #d#r##]`
At the end of the each sub-island, we put a number sign to indicate that the sub-island's DFS has come to an end.  We do not need open parentheses as just showing the end with "#" is enough. This way, it generates unique island codes for each island. Trying to make sense of these encodings by comparing it with the grid is the best way to understand this.

altayhunoglu (approved):
Thank you for your valuable collaboration. It was really a nice question to be more familiar with DFS.

ErdemT09:
I will actually add the other solution with coordinates to this PR, it is nonetheless more straightforward than this, as this one resembles some esoteric programming language  like [this](https://en.wikipedia.org/wiki/Unlambda) or this [this](https://en.wikipedia.org/wiki/Brainfuck).

ErdemT09:
While working on the solution with the Cartesian coordinate differences pairs, I have come across a small discovery or an issue about how Java treats lists of arrays.
For this grid:
```
grid = new int[][] { 
            { 1, 1, 1 },
            { 0, 1, 0 }, 
            { 0, 1, 0 },
            { 0, 0, 1 }, 
            { 0, 0, 1 },
            { 1, 1, 0 },
            { 1, 0, 1 },
            { 0, 1, 1 }, 
            { 1, 0, 0 }, 
            { 1, 0, 0 }, 
            { 0, 1, 0 } };
```
The program should store the list of islands as set of şlist of arrays (of length 2). If two lists of arrays are the same, according to us, it should also be same for Java, I though. But it is apparently not.
The islands are apparently this:
`[ {[0, 0][1, 0]}, {[0, 0][1, 0][0, 1]}, {[0, 0][1, 0][1, -1]}, {[0, 0]}, {[0, 0][1, 0]}, {[0, 0][0, 1][1, 1][2, 1][0, 2]} ]`
The two vertical 2 length blocks weren't considered identical by Java. This probably has to do with the discretion between Java collections and Java arrays.

Solution to this problem is using a set of list of lists.

ErdemT09:
Are they both ready to be merged now?

src/main/java/algorithms/curated170/medium/numberofdistinctislands/NumberOfDistinctIslandsCoordinates.java
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    islands.add(currIsland);
+                }
+            }
+        }
+
+        return islands.size();
+    }
+
+    int[] parent;
```
altayhunoglu, line 36: With "parent", it takes 8ms.
Without it, takes 6ms.
Like this:
```
public class Solution {

    Set<String> islands;
    int[][] grid; 
     
    int xDim;
    int yDim;
    public int numDistinctIslands(int[][] grid)
    {
        this.grid = grid; 
        islands = new HashSet<>();
        xDim = grid.length;
        yDim = grid[0].length;

        for(int i = 0; i<xDim; i++)
        {
            for(int j = 0; j<yDim; j++)
            {
               islandCode = new StringBuilder();
                 DFS(i, j, "#");
                  if (islandCode.length() == 0) {
                    continue;
                }
                islands.add(islandCode.toString());
            }
        }

        return islands.size();
    }
    StringBuilder islandCode = new StringBuilder();
    private void DFS(int i, int j, String dir)
    {
      if (i < 0 || j < 0 || i >= grid.length || j >= grid[0].length || grid[i][j] == 0) return;    
            
          grid[i][j] = 0;
       
           islandCode.append(dir);
            DFS(i+1, j, "d");
            DFS(i-1, j, "u");
            DFS(i, j+1, "r");
            DFS(i, j-1, "l");
            islandCode.append("#");
    
    }

 
}
```

ErdemT09:
It works.

altayhunoglu (approved):
Great code. Thanks.

ErdemT09:
We have to delete that temporary file but I will actually do it from master.

PR #198, opened 2021-05-12
249-Small refactoring suggestion

ErdemT09:
Creating a new variable called list seemed unnecessary, so I replaced it with other HashMap methods.


altayhunoglu (approved):
Great point Erdem. Thanks.

Issue #186, opened 2021-05-09
1061. Lexicographically Smallest Equivalent String


altayhunoglu:
Varsayılan kod:
```
class Solution {
    public String smallestEquivalentString(String A, String B, String S) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/117588155-5653b000-b12a-11eb-94fa-c17aea441a7e.png)
![image](https://user-images.githubusercontent.com/60903744/117588165-69668000-b12a-11eb-8384-ea2bf98f69e6.png)
![image](https://user-images.githubusercontent.com/60903744/117588169-6ff4f780-b12a-11eb-8879-9dcf89368634.png)

altayhunoglu:
solved:
https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/medium/LexicographicallySmallestEquivalentString.java

PR #199, opened 2021-05-12
1102. Lexicographically Smallest Equivalent String Methods Patch

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/186)
Firstly, the solution procedure was in the main method, so I created a method as wanted in the question an moved the solution there.

Additionally, in the find method we have to add a `getOrDefault( )` check. In our target String `S`, there might be letters that were not present in `A` and `B`. We do not replace these letters and let them stay as they are, because the equivalence operation is reflexive here. The reason why this is important is understandable in the example `smallestEquivalentString("leetcode", "programs", "sourcecode")`
There, the output is `"aauaaaaada"`, the 'u' is left unchanged, because we haven't encountered it in the two strings, which means that its smallest equivalent char is itself.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/LexicographicallySmallestEquivalentString.java
+++ b/src/main/java/algorithms/curated170/medium/LexicographicallySmallestEquivalentString.java
@@ -7,7 +7,7 @@ public class LexicographicallySmallestEquivalentString {
 
     private Map<Character, Character> parent;
 
-    String smallestEquivalentString(String A, String B, String S) {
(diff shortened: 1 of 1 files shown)


ErdemT09:
https://github.com/nataliekung/leetcode/blob/master/dfs/lexicographically-smallest-equivalent-string.md
Also, is the DFS solution here anyhow more efficient than the current solution?
Even if not, I think we should translate it to Java.

altayhunoglu (approved):
You are right. Thanks.

PR #200, opened 2021-05-12
323. Number of Connected Components in an Undirected Graph

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/31)
There are 2 solutions for this question, similar to [this](https://github.com/spiralgo/algorithms/issues/186), a Union-Find approach, and a DFS approach.

Approach 1: DFS:
This is a graph question and the graph is pretty easy to construct in this question. For this, we can use a list of lists as a graph, where we add both corners of an edge to their respective lists in the graph, as it should be undirected.
After that, we can just search through each node and their neighbors using DFS. We mark all nodes we encounter as visited, so we don't DFS them or their neighbors again. While looping through our graph list, if we can encounter a non-visited node, that means that there is an unvisited "connected component" here. We increment our counter according to this.
In this image, for example:
image
We begin iterating from loop 0 and DFS this node as it is not visited. 
In DFS, we mark it visited.
From its neighbors, we first visit 1 and mark it visited. 0 is a neighbor of 1, but we return back when we visit it because it is already visited. From 1's neighbors, we first visit 2 and mark it visited. At 2, We return back when we visit 0 and 1 again.
Then we DFS to 1's other neighbor 8 and to 5 from there. Same visited checks also happen here. 
While iterating through the graph, when try to DFS the node 2 for example, the procedure will immediately return as 2 is marked visited.
The same DFS will also take place by 3 and 4. We can check if our DFS is successful by assigning it a integer value of 0 and 1.

Approach 2: Union-find:
It is important to understand that:
If a new given edge connects 2 previously discrete nodes, then the number of connected components decreases.
If we have [0,1] and  [2,3], we have 4 nodes, and 2 edges reduces the number of connected components to 2.
If we connect 2 and 1 for example, the number will reduce to 1.
We can check in which component a node is by storing its root node somewhere. Until a node is in its own root, we have to search recursively through  a node's root, in order to find the main root. And set one's root to be the another.
At the end, we return our reduced number of nodes as to be seen in the code.


ErdemT09:
Approach 3: BFS:
We just do the same that we had done with DFS, but we visit nodes layer by layer, avoiding recursion.

src/main/java/algorithms/curated170/medium/numberofconnectedcomponents/NumberOfConnectedComponentsDFS.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class NumberOfConnectedComponentsDFS {
+
+    boolean[] visited;
+    List<List<Integer>> connections = new ArrayList<>();
+
+    public int countComponents(int[][] edges, int n) {
```
altayhunoglu, line 11: Very nice solution.
Can we alter the params order:
   public int countComponents(int n, int[][] edges) 
   
   
image

src/main/java/algorithms/curated170/medium/numberofconnectedcomponents/NumberOfConnectedComponentsUnionFind.java
```diff
@@ -0,0 +1,51 @@
...
+        return n;
+    }
+
+    private int findRoot(int key) {
+        if (key == roots[key])
+        {
+            return key;
+        }
+        return findRoot(roots[key]);
+    }
```
altayhunoglu, line 36: If we use "path compression" it will take 1ms; otherwise, it takes 4ms.
 
     ```
private int findRoot(int key) {
        if (key != roots[key])
        {
            roots[key] = roots[roots[key]];
            key = roots[key];
            return findRoot(key);
        }else{
            return key;
        }
    }
```

File: src/main/java/algorithms/curated170/medium/numberofconnectedcomponents/NumberOfConnectedComponentsUnionFind.java [hunk 3]
```diff
@@ -0,0 +1,51 @@
+package algorithms.curated170.medium.numberofconnectedcomponents;
+
+public class NumberOfConnectedComponentsUnionFind {
+
+    int roots[];
+
+    public int countComponents(int[][] edges, int n) {
```
altayhunoglu: Can we alter the params order:
public int countComponents(int n, int[][] edges)

ErdemT09: All fixed.

ErdemT09: All fixed.

altayhunoglu (approved):
All approved.
 Thanks for these nice solutions and drawing.

ErdemT09:
How do the other two solutions perform on LeetCode?

altayhunoglu:
> How does the BFS solution perform on LeetCode? It shouldn't be different from DFS though.

Yes. It is fine enough with 4ms.

PR #201, opened 2021-05-13
1061. Lexicographically Smallest Equivalent String-DFS

ErdemT09:
We could think of the equivalence relation between letters as a graph.
For each letter in the target String, we can DFS through this graph to find the smallest char value that is in a connected component of the graph with this letter.
That is the idea behind this solution.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/LexicographicallySmallestEquivalentStringDFS.java
+++ b/src/main/java/algorithms/curated170/medium/LexicographicallySmallestEquivalentStringDFS.java
@@ -1,6 +1,7 @@
 package algorithms.curated170.medium;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Nice solution.
It took, 59ms. I think it is because of the HashSet.

I also tried alternative ways to measure the efficiency, you can choose which one of those three you like.

When I try with a boolean array for "seen" instead, it takes 38ms.
(With a little refactoring on DFS)
```

class Solution {
private Map<Character, List<Character>> parent;
    boolean[] seen;
    public String smallestEquivalentString(String A, String B, String S) {
        parent = new HashMap<>();

        for (int i = 0; i < A.length(); i++) {
            parent.putIfAbsent(A.charAt(i), new ArrayList<>());
            parent.putIfAbsent(B.charAt(i), new ArrayList<>());

            parent.get(A.charAt(i)).add(B.charAt(i));
            parent.get(B.charAt(i)).add(A.charAt(i));
        }

        StringBuilder sb = new StringBuilder("");
        for (char c : S.toCharArray()) {
            seen = new boolean[26];
            sb.append(DFS(c,c));
        }
        return sb.toString();
    }

    private char DFS(char c, char mn)
    {
        if(seen[c-'a'])
        {
            return mn;
        }
        seen[c-'a'] = true;
        char res = mn;
        List<Character> w = parent.getOrDefault(c, Collections.emptyList());
        for(char n : w) //or List.of()  or new ArrayList<>()
        {
           
                  mn = seen[n-'a'] ? mn : (char) DFS(n, mn);
             
        }
        return  (char) Math.min(res, (char) Math.min(mn, c));
    }
}
```


If we use an array instead of HashMap, it takes 28ms:

```
class Solution {
private  List<Character>[] parent;
    boolean[] seen;
    public String smallestEquivalentString(String A, String B, String S) {
        parent = new ArrayList[26];

        for (int i = 0; i < A.length(); i++) {
            int a = A.charAt(i)-'a';
            int b = B.charAt(i)-'a';
            if(parent[a] == null){
                parent[a] = new ArrayList<>();
            }
             if(parent[b] == null){
                parent[b] = new ArrayList<>();
            } 

            parent[a].add(B.charAt(i));
            parent[b].add(A.charAt(i));
        }

        StringBuilder sb = new StringBuilder("");
        for (char c : S.toCharArray()) {
            seen = new boolean[26];
            sb.append(DFS(c,c));
        }
        return sb.toString();
    }

    private char DFS(char c, char mn)
    {
        if(seen[c-'a'])
        {
            return mn;
        }
        seen[c-'a'] = true;
        char res = mn;
        List<Character> charList = parent[c-'a'] == null ? new ArrayList<>() :  parent[c-'a']; 
        for(char n : charList) //or List.of()  or new ArrayList<>()
        {
          mn = seen[n-'a'] ? mn : (char) DFS(n, mn);
        }
        return  (char) Math.min(res, (char) Math.min(mn, c));
    }
}
 

```

ErdemT09:
Additionally, I have added some constants and a small method for getting the index of char. This makes the code more modular.

ErdemT09:
We might not actually need the variable `res`:
```
private char DFS(char c, char mn) {

        int charIndex = letterIndex(c);

        if (seen[charIndex]) {
            return mn;
        }
        seen[charIndex] = true;
        
        List<Character> charList = parent[charIndex] == null ? Collections.emptyList() : parent[charIndex]; // or List.of() or new ArrayList<>(). I prefer this, because it is more readable.
        for (char n : charList) 
        {
            mn = seen[letterIndex(n)] ? mn : (char) DFS(n, mn);
        }
        return (char) Math.min(c, mn);
    }
```

altayhunoglu (commented):
> We might not actually need the variable `res`:
> 
> ```
> private char DFS(char c, char mn) {
> 
>         int charIndex = letterIndex(c);
> 
>         if (seen[charIndex]) {
>             return mn;
>         }
>         seen[charIndex] = true;
>         
>         List<Character> charList = parent[charIndex] == null ? Collections.emptyList() : parent[charIndex]; // or List.of() or new ArrayList<>(). I prefer this, because it is more readable.
>         for (char n : charList) 
>         {
>             mn = seen[letterIndex(n)] ? mn : (char) DFS(n, mn);
>         }
>         return (char) Math.min(c, mn);
>     }
> ```

You are absolutely right. We can remove res. It takes 27ms.

We could also remove ternary if-statement like this.
```
        for(char n : charList) //or List.of()  or new ArrayList<>()
        {
          mn = DFS(n, mn);
        }
```
 But, it takes 32ms. I think it is about the extra method call.

src/main/java/algorithms/curated170/medium/LexicographicallySmallestEquivalentStringDFS.java
```diff
@@ -17,8 +14,8 @@ public String smallestEquivalentString(String A, String B, String S) {
         parent = new ArrayList[26];
 
         for (int i = 0; i < A.length(); i++) {
-            int a = A.charAt(i) - 'a';
-            int b = B.charAt(i) - 'a';
+            int a = letterIndex(A.charAt(i));
+            int b = letterIndex(B.charAt(i));
```
altayhunoglu, line 18: It is good to remove print statements inside the business code as it slows down the run-time.
 The future audience may misvalue the efficiency if they test without removing the print part.

ErdemT09: I should be more careful about not forgetting them.

altayhunoglu (approved):
Thank you for your valuable collaboration.

PR #202, opened 2021-05-13
244. Shortest Word Distance II

ErdemT09:
At first glance, this question looks like just a re-writing of [Shortest Word Distance I](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/easy/ShortestWordDistance.java). However, it's actually deeper than that.
While initializing the Class, we can actually store the indices of the words in Map of <String, List<Integer>>. 
After that, comparing two words is basically same as:
Find the smallest distance between two numbers given two sorted list of numbers.
We check, update the minimum distance each time. If the second number is greater than the first, we increment our first index, else we increment our second index. 

I am not sure if the given two words can be same. If not, then we can delete the 0-distance check.

I have also removed an unused variable from the previous question.

### DIFF
--- a/src/main/java/algorithms/curated170/easy/ShortestWordDistance.java
+++ b/src/main/java/algorithms/curated170/easy/ShortestWordDistance.java
@@ -3,8 +3,7 @@ package algorithms.curated170.easy;
 public class ShortestWordDistance {
   public int shortestDistance(String[] words, String word1, String word2) {
     int i1 = -1, i2 = -1;
-    int minDistance = words.length;
-    int currentDistance;
+    int minDist = words.length;
     for (int i = 0; i < words.length; i++) {
         if (words[i].equals(word1)) {
             i1 = i;
@@ -13,10 +12,10 @@ public class ShortestWordDistance {
         }
 
         if (i1 != -1 && i2 != -1) {
-            minDistance = Math.min(minDistance, Math.abs(i1 - i2));
+            minDist = Math.min(minDist, Math.abs(i1 - i2));
         }
     }
-    return minDistance;
+    return minDist;
 }
   
 }
--- a/src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
+++ b/src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
@@ -24,7 +24,7 @@ public class ShortestWordDistance2 {
         public int shortest(String word1, String word2) {
             List<Integer> ind1 = wordIndices.get(word1);
             List<Integer> ind2 = wordIndices.get(word2);
-            int minDist = wordIndices.size();
+            int minDist = Integer.MAX_VALUE;
             int i = 0, j = 0;
             while (i < ind1.size() && j < ind2.size()) {
                 int i1 = ind1.get(i);
--- a/src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
+++ b/src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
@@ -1,0 +1,52 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+public class ShortestWordDistance2 {
+
+    public class WordDistance {
+
+        HashMap<String, List<Integer>> wordIndices;
+
+        public WordDistance(String[] words) {
+            wordIndices = new HashMap<String, List<Integer>>();
+
+            for (int i = 0; i < words.length; i++) {
+                if (!wordIndices.containsKey(words[i])) {
+                    wordIndices.put(words[i], new ArrayList<Integer>());
+                }
+                wordIndices.get(words[i]).add(i);
+            }
+        }
+
+        public int shortest(String word1, String word2) {
+            List<Integer> ind1 = wordIndices.get(word1);
+            List<Integer> ind2 = wordIndices.get(word2);
+            int minDist = wordIndices.size();
+            int i = 0, j = 0;
+            while (i < ind1.size() && j < ind2.size()) {
+                int i1 = ind1.get(i);
+                int i2 = ind2.get(j);
+                int currDist = i1 - i2;
+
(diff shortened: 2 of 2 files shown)


src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
```diff
@@ -0,0 +1,52 @@
...
+            int i = 0, j = 0;
+            while (i < ind1.size() && j < ind2.size()) {
+                int i1 = ind1.get(i);
+                int i2 = ind2.get(j);
+                int currDist = i1 - i2;
+
+                minDist = Math.min(minDist, Math.abs(currDist));
+
+                if (currDist == 0) {
+                    return 0;
```
ErdemT09, line 37: Removing this makes it work. 
But why?

ErdemT09:
Do we need to change anything with Shortest Word Distance ( I )?

src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
altayhunoglu: Yeah. It is most probably related to this:
image

src/main/java/algorithms/curated170/medium/ShortestWordDistance2.java
```diff
@@ -0,0 +1,55 @@
...
+                    wordIndices.put(words[i], new ArrayList<Integer>());
+                }
+                wordIndices.get(words[i]).add(i);
+            }
+        }
+
+        public int shortest(String word1, String word2) {
+            List<Integer> ind1 = wordIndices.get(word1);
+            List<Integer> ind2 = wordIndices.get(word2);
+            int minDist = wordIndices.size();
```
altayhunoglu, line 27: We should change this. Because "two words" will constantly change at every **shortest(String word1, String word2)** call. It is why we have to reset the **minDist** too. We can do it using **words.length**, but this is easier:
            
```suggestion
           int minDist = Integer.MAX_VALUE;
```

src/main/java/algorithms/curated170/easy/ShortestWordDistance.java
```diff
@@ -3,8 +3,7 @@
 public class ShortestWordDistance {
```
altayhunoglu, line 3: We should change like this. But, it is not an efficient way, as far as I understand.
Because the "shortest" method call will be done too many times. 
image

The other approach uses precomputation. It is why it was more appropriate for this case.
But maybe we can find a way to handle with this approach too.

 
```
class WordDistance {
    String[] words;
    public WordDistance(String[] wordsDict) {
         this.words = wordsDict;
    }
    
    public int shortest(String word1, String word2) {
      
        int minDist = words.length, i1 = minDist, i2 = -minDist;
        for (int i = 0; i < words.length; i++) {
        if (words[i].equals(word1)) {
            i1 = i;
        } else if (words[i].equals(word2)) {
            i2 = i;
        }

       
            minDist = Math.min(minDist, Math.abs(i1 - i2));
        
    }
    return minDist;
    }
     
}

```

ErdemT09: Wasn't the original question just about creating a method and not a class?

altayhunoglu: I updated the issue with images and the default code.
Sorry for the late consideration.
https://github.com/spiralgo/algorithms/issues/42

ErdemT09: I don't think looping through all the indices of the words array can be or be improved to be efficient like the map solution.

altayhunoglu: Yeah, you are right.
Maybe for the nonce, we can retreat and remove this approach until the time we find an efficient way.
As you know, sometimes we should lose the battles to win the war.

ErdemT09: What should we upload now then?

altayhunoglu: The other approach is working quite fine.

ErdemT09: Are we done with that approach or should we change anything further?

altayhunoglu: I mean we can remove ShortestWordDistance.java until we can find an efficient way to do this without precomputation (at least enough to pass the test cases.)

The precomputation approach works nice.

ErdemT09: > I mean we can remove ShortestWordDistance.java until we can find an efficient way to do this without precomputation (at least enough to pass the test cases.)
> 
> The precomputation approach works nice.
> 

The other file is the easy version of the question without classes.

altayhunoglu: But it works?
image

altayhunoglu (approved):

src/main/java/algorithms/curated170/easy/ShortestWordDistance.java
altayhunoglu: The question is this one, right?
https://leetcode.com/problems/shortest-word-distance-ii/submissions/

ErdemT09: Yes.
However, I have also added a small change to [this question](https://github.com/spiralgo/algorithms/issues/147). That is the file that you have started this conversation in.

ErdemT09: I shall merge then if you have also understood the situation.

altayhunoglu: Oke, we can merge. Thank you, Erdem.

PR #203, opened 2021-05-13
348. Design Tic-Tac-Toe 

ErdemT09:
[Issue.](https://github.com/spiralgo/algorithms/issues/37)
In this game, there are 4 ways to win:
In an n by n grid:
1-Fill an entire row
2-Fill an entire column
3-Fill the entire leading diagonal
4-Fill the entire counterdiagonal

Instead of checking if a given coordinate tile has tiles that satisfy a victory, we can check if a player has n markings in the row, column or perhaps the diagonal corresponding to this tile.

If none of the 4 conditions are met, we just return false { 0 }.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
+++ b/src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
@@ -1,0 +1,49 @@
+package algorithms.curated170.medium;
+
+import java.util.Arrays;
+import java.util.Random;
+
+public class DesignTicTacToe {
+
+    public class TicTacToe {
+        int[][] countRow;
+        int[][] countCol;
+        int[] countDia1;
+        int[] countDia2;
+        int length;
+
+        public TicTacToe(int n) {
+            length = n;
+            countRow = new int[n][3];
+            countCol = new int[n][3];
+            countDia1 = new int[3];
+            countDia2 = new int[3];
+        }
+
+        public int move(int row, int col, int player) {
+
+            if (countRow[row][player]++ == length)
+                return player;
+
+            if (countCol[col][player]++ == length)
+                return player;
+
+            if (row == col) {
+
+                if (countDia1[player] == length)
+                    return player;
+            }
+
+            if (row + col == length - 1) {
+                if (countDia2[player] == length)
+                    return player;
+            }
+
+            return 0; // No one wins
+        }
+    }
+
+    public static void main(String[] args) {
+
+    }
+}
--- a/src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
+++ b/src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
@@ -1,8 +1,5 @@
 package algorithms.curated170.medium;
 
-import java.util.Arrays;
-import java.util.Random;
-
 public class DesignTicTacToe {
 
     public class TicTacToe {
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
```diff
@@ -0,0 +1,49 @@
...
+            countRow = new int[n][3];
+            countCol = new int[n][3];
+            countDia1 = new int[3];
+            countDia2 = new int[3];
+        }
+
+        public int move(int row, int col, int player) {
+
+            if (countRow[row][player]++ == length)
+                return player;
```
altayhunoglu, line 26: I think you meant:
` if (++countRow[row][player] == length)`

src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
```diff
@@ -0,0 +1,49 @@
...
+
+            if (countCol[col][player]++ == length)
```
altayhunoglu, line 28: I think you meant:
++before

src/main/java/algorithms/curated170/medium/DesignTicTacToe.java
```diff
@@ -0,0 +1,49 @@
...
+                return player;
+
+            if (row == col) {
+
+                if (countDia1[player] == length)
```
altayhunoglu, line 33: I think you meant:
++increment

altayhunoglu:
When we fix the increments, it works like a charm and takes 3ms.
  
  image

  
   We can also reduce if-statements even if it will not affect the efficiency, it looks lean and minimalistic. You can choose.
   
   ```
public class TicTacToe {
        int[][] countRow;
        int[][] countCol;
        int[] countDia1;
        int[] countDia2;
        int length;

        public TicTacToe(int n) {
            length = n;
            countRow = new int[n][3];
            countCol = new int[n][3];
            countDia1 = new int[3];
            countDia2 = new int[3];
        }

        public int move(int row, int col, int player) {
          
         return (++countRow[row][player] == length 
             ||  ++countCol[col][player] == length
             ||  row == col && ++countDia1[player] == length
             ||  row + col + 1 == length &&  ++countDia2[player] == length
                ) ? player : 0; 
   
        }
    }

```

File: src/main/java/algorithms/curated170/medium/DesignTicTacToe.java (hunk 3 above)
Comment (ErdemT09) [reply src/main/java/algorithms/curated170/medium/DesignTicTacToe.java:33]: I didn't even know their difference and thought it was stylistic. Thanks for the information.

File: src/main/java/algorithms/curated170/medium/DesignTicTacToe.java [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+            countDia1 = new int[3];
+            countDia2 = new int[3];
+        }
+   
+        public int move(int row, int col, int player) {
+          
+         return ((++countRow[row][player] == length) 
+             ||  (++countCol[col][player] == length)
+             ||  (row == col && ++countDia1[player] == length)
+             ||  (row + col + 1 == length &&  ++countDia2[player] == length)
```
ErdemT09, line 25: I have added parenthesis around these statements to make them look a clearer. It will also speed up the code by a really tiny, almost negligible fraction, but in diagonals, the method won't try to look up the array's value if row is not equal to the column or the other possibility.

altayhunoglu (approved):
Nice and clear code.

PR #204, opened 2021-05-14
136. and 137. (Single number I and II)

ErdemT09:
136: [Question.](https://leetcode.com/problems/single-number/)
Problem restatement:
We are given an array of integers. There, every number appears twice, except one of them, which only appears once.
Find and return that number. 
"You must implement a solution with a linear runtime complexity and use only constant extra space."

Solution:
How would a human would solve this problem? Of course, note the number of times a number appears somewhere. After checking the entire array, find the value corresponding to count == 1 from the noted values. This is equivalent to using a HashMap or a large array.
But logic and computer bits allow us a more efficient solution.

Let S be an empty set, where we will store our numbers.
If we XOR a number with this set, it will add the number if it's absent, or remove the number if it's present. 
This operation for sets is also valid for binary numbers.
`0 ^ 1 ^ 1 = 0 ^ (1 ^1) = 0 ^ 0 = 0`, a set's XOR with itself is 0
From this, we can infer that xor'ing all the numbers in the array would yield 0 ^ unique, which equals to unique:
```
Input: [ 2, 1, 4, 9, 4, 2, 1 ]
Output: 0 ^ 2 ^ 1  ^ 4 ^ 9 ^ 4^2 ^ 1 =  0 ^ (1 ^1) ^ (2 ^2) ^ (4 ^4)  ^ 9 = 0 ^ 9 = 9
```

137: [Question. ](https://leetcode.com/problems/single-number-ii/)

This is a bit more complicated:
For this question, we need 2 sets:
S1: Storing numbers that have appeared once
S2: Storing numbers that have appeared twice
If a number is not in S1 and S2 we add it to S1, if a number is in S1 but not S2, we should remove it from S1 and add it to S2.
Same is also valid for S2. If a number is not in S1 (or has been removed from it) and it is not in S2, we add it. Else, we remove it from here, which means that the number has appeared thrice and we are done counting it, kind of like a clock. 

This way, numbers that are in S1 and not in S2 pop to S2 if encountered again.
Numbers in S2 to get removed if not in S1 and are in S2 already.
This way all numbers that occur 3 times get removed from both sets and only the unique number is left in S1.
We should return S1.
This corresponds to the following binary operation:
```
x1 = (x1 ^ num) & ~x2;
x2 = (x2 ^ num) & ~x1;
```

We can generalize this to 4 or larger number of appearances of integers:
```

            x1 = (x1 ^ num) & ~x2 & ~x3;
            x2 = (x2 ^ num) & ~x1 & ~x3;
            x3 = (x3 ^ num) & ~x1 & ~x2;
```
Which is equivalent to saying: For each set, If the number is not in this set and it is not in other sets, add it to the set. Else, if in the set, remove it from the set.

I hope I have made the solution and the generalization clear. I was for some reason unable to write the code that generalizes this expression, it does not work for some reason.

### DIFF
--- a/src/main/java/algorithms/easy/SingleNumber.java
+++ b/src/main/java/algorithms/easy/SingleNumber.java
@@ -1,4 +1,4 @@
-package algorithms.medium;
+package algorithms.easy;
 
 public class SingleNumber {
 
--- a/src/main/java/algorithms/medium/SingleNumber.java
+++ b/src/main/java/algorithms/medium/SingleNumber.java
@@ -1,0 +1,22 @@
+package algorithms.medium;
+
+public class SingleNumber {
+
+    public int singleNumber(int[] nums) {
+        int x1 = 0;
+        for (int n : nums) {
+            x1 ^= n;
+        }
+        return x1;
+    }
+
+    public static void main(String[] args) {
+        var solution = new SingleNumber();
+
+        int[] nums = new int[] { 2, 2, 3, 5, 7, 3, 9, 5, 7 };
+        int[] nums2 = new int[] { 4, 3, 6, 0, 6, 4, 3 };
+
+        System.out.println(solution.singleNumber(nums)); // prints 9
+        System.out.println(solution.singleNumber(nums2)); // prints 0
+    }
+}
--- a/src/main/java/algorithms/medium/SingleNumber2.java
+++ b/src/main/java/algorithms/medium/SingleNumber2.java
@@ -1,0 +1,98 @@
+package algorithms.medium;
+
+public class SingleNumber2 {
+
+    public int singleNumber(int[] nums) {
+
+        int x1 = 0, x2 = 0;
+
+        for (int num : nums) {
+
+            x1 = (x1 ^ num) & ~x2;
+            x2 = (x2 ^ num) & ~x1;
+        }
+        return x1;
+    }
+
+    public int singleNumber4(int[] nums) {
+
+        int x1 = 0, x2 = 0, x3 = 0;
+
+        for (int num : nums) {
+
+            x1 = (x1 ^ num) & ~x2 & ~x3;
+            x2 = (x2 ^ num) & ~x1 & ~x3;
+            x3 = (x3 ^ num) & ~x1 & ~x2;
+        }
+        return x1;
+    }
+
+    public int singleNumber5(int[] nums) {
+
+        int x1 = 0, x2 = 0, x3 = 0, x4 = 0;
+
+        for (int num : nums) {
+
+            x1 = (x1 ^ num) & ~x2 & ~x3 & ~x4;
+            x2 = (x2 ^ num) & ~x1 & ~x3 & ~x4;
+            x3 = (x3 ^ num) & ~x1 & ~x2 & ~x4;
+            x4 = (x4 ^ num) & ~x1 & ~x2 & ~x3;
+        }
+        return x1;
+    }
+
+    public int singleNumber6(int[] nums) {
+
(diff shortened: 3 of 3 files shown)


ErdemT09:
https://leetcode.com/problems/sum-of-two-integers/discuss/84278/A-summary:-how-to-use-bit-manipulation-to-solve-problems-easily-and-efficiently
Noting this here for the future.

src/main/java/algorithms/easy/SingleNumber.java
```diff
@@ -0,0 +1,22 @@
+package algorithms.easy;
+
+public class SingleNumber {
+
+    public int singleNumber(int[] nums) {
+        int x1 = 0;
+        for (int n : nums) {
+            x1 ^= n;
+        }
+        return x1;
```
altayhunoglu, line 10: Quite efficient.

image

altayhunoglu:
Very nice idea to exercise bitwise operations with this generalization.
 Thank you for bringing it out.
 
 I think we can think like this:

 So if k = 3:
```
x1 = (x1 ^ num) & ~x2;
x2 = (x2 ^ num) & ~x1;
```
We can generalize it like this:

If k = i + 2
```
x[i] = (x[i] ^ num) & ~x[i+1];
x[i+1] = (x[i+1] ^ num) & ~x[i];
```
 
 If we implement it like this:
 
```
  public int singleNumber(int[] nums, int k) {
       
        int[] x = new int[k];
 
        for (int num : nums) {
           for (int i = 1; i < k-1; i++) {
                 x[i] = (x[i] ^ num) & ~x[i+1];
                 x[i+1] = (x[i+1] ^ num) & ~x[i];
           }
        }
        return x[1];
    }
```

And we can test the generalized version over the specialized one like this:
image

altayhunoglu:
Very nice idea to exercise bitwise operations with this generalization.
 Thank you for bringing up the bit operations title.
 
 I think we can think like this:

 **So if k = 3:**
```
x1 = (x1 ^ num) & ~x2;
x2 = (x2 ^ num) & ~x1;
```
We can generalize it like this:

 ```
 public int singleNumber(int[] nums, int k) {
       
       int x1 = 0, x2 = 0;
       HashMap<Integer, Integer> map = new HashMap<>();
        for (int num : nums) {
            map.merge(num,1, Integer::sum);
            if(map.get(num)>3) continue;
            x1 = (x1 ^ num) & ~x2;
            x2 = (x2 ^ num) & ~x1; 
           
        }
        return x1;
    }


```

And we can test the generalized version over the specialized one like this:
image

ErdemT09:
> Very nice idea to exercise bitwise operations with this generalization.
> Thank you for bringing up the bit operations title.
> 
> I think we can think like this:
> 
> **So if k = 3:**
> 
> ```
> x1 = (x1 ^ num) & ~x2;
> x2 = (x2 ^ num) & ~x1;
> ```
> 
> (As a temporary solution until we find a pure bitwise way) We can generalize it like this:
> 
> ```
> public int singleNumber(int[] nums) {
>       
>       int x1 = 0, x2 = 0;
>       HashMap<Integer, Integer> map = new HashMap<>();
>        for (int num : nums) {
>            map.merge(num,1, Integer::sum);
>            if(map.get(num)>3) continue;
>            x1 = (x1 ^ num) & ~x2;
>            x2 = (x2 ^ num) & ~x1; 
>           
>        }
>        return x1;
>    }
> ```

Solved this problem. We use n-1 sets when there n occurrences of numbers.

altayhunoglu (approved):
The generalized version also works like a charm:

image

PR #205, opened 2021-05-15
1055. Shortest Way to Form String

ErdemT09:
This question is a fairly simple two pointers problem.
The target is composed of substrings that are subsequences of the source string. From this, we can infer the solution that we should compare characters of the target and the source in some way. We can do this by looping through the source string each time and checking if the given char at p1 is the same as the one in the target at p2. We should keep track of p2 somewhere until it is bigger than the indices of the target string, if so, we should return how many times we have looped through the original string, which corresponds to how many subsequences of the source we have checked. 
While iterating through the source string, if p2 does not move anywhere, that means that target contains a char at p2 that p1 does not contains, which means that target cannot be constructed from the source string. We can check this with a simple variable.
The image below illustrates this approach:
image


altayhunoglu (approved):
Thank you for your beautiful explanation.
image

PR #206, opened 2021-05-15
1426. Counting Elements-Two alternative solutions

ErdemT09:
We can actually use the constraints given in the question to store the existence of numbers in a boolean array.  The array is looped once again and we check if number+1 is true in the array. I personally do not really like this solution as it wouldn't be valid if the constraints were something unknown to us and not limited by LeetCode or the programming language itself (For example, `Integer `type in Haskell represents an unbounded integer value).
The other solution is a bit more general, and can be pretty fast as well depending on the implementation of the HashSet. We add all the numbers in the array to a set and after that, check if the sought value is in the set.

### DIFF
--- a/src/main/java/algorithms/curated170/easy/CountingElementsLinear.java
+++ b/src/main/java/algorithms/curated170/easy/CountingElementsLinear.java
@@ -1,0 +1,48 @@
+package algorithms.curated170.easy;
+
+public class CountingElementsLinear {
+    public int countElements(int[] arr) {
+        boolean[] check = new boolean[1002];
+        int count = 0;
+        for (int n : arr) {
+            check[n] = true;
+        }
+        for (int n : arr) {
+            if (check[n + 1]) {
+                count++;
+            }
+        }
+        return count;
+    }
+
+    public static void main(String[] args) {
+
+        var solution0 = new CountingElementsLinear();
+        var solution1 = new CountingElements();
+
+        int[] arr = { 1, 2, 3 };
+        System.out.println(solution0.countElements(arr));
+        System.out.println(solution1.countElements(arr));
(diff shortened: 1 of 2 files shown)


altayhunoglu:
issue: https://github.com/spiralgo/algorithms/issues/150

src/main/java/algorithms/curated170/easy/CountingElementsLinear.java
```diff
@@ -0,0 +1,48 @@
...
+            check[n] = true;
+        }
+        for (int n : arr) {
+            if (check[n + 1]) {
+                count++;
+            }
+        }
+        return count;
+    }
+
```
altayhunoglu, line 17: It is quite fine.
image

src/main/java/algorithms/curated170/easy/CountingElementsSet.java
```diff
@@ -0,0 +1,52 @@
...
+        for (int n : arr) {
+            nums.add(n);
+        }
+        for (int n : arr) {
+            if (nums.contains(n + 1)) {
+                count++;
+            }
+        }
+        return count;
+    }
```
altayhunoglu, line 20: This is quite fine as well.
image

altayhunoglu (approved):
Thank you for these nice solutions Erdem.

src/main/java/algorithms/curated170/easy/CountingElementsSet.java
ErdemT09: How does the current solution with sorting perform?

altayhunoglu: image

PR #207, opened 2021-05-15
1060. Missing Element in Sorted Array 

ErdemT09:
Issue: #38 
The first approach is a brute-force approach. We check whether the number of missing numbers remained is in the range of difference of two numbers, if not, we subtract this range from the amount of missing numbers and eventually reach the solution. This is however a linear time algorithm. We can improve this to logarithmic time using Binary Search.

We can calculate the number of missing numbers at some index i using the following method:
```
 private int countMissing(int[] nums, int i) {
        return nums[i] - nums[0] - i;
    }
```
Let's say we have the array {4, 10}. The missing numbers at index 1 should be 10 - 4 - 1 = 5, which is true and should give the intuition about why this method works. 
With this method, we can binary search which index is appropriate to this method. At the end, we return the value at this index added with the order of missing number minus the missing numbers that have previously occurred before this index.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/MissingElement.java
+++ b/src/main/java/algorithms/curated170/medium/MissingElement.java
@@ -2,20 +2,21 @@ package algorithms.curated170.medium;
 
 public class MissingElement {
 
-    public int missingElement(int[] nums, int k) {
-        int l = 0;
-        int r = nums.length;
+       public int missingElement(int[] nums, int k) {
+        int left = 0;
+        int right = nums.length;
 
-        while (l < r) {
-            int m = l + (r - l) / 2;
+        while (left < right) {
+            int m = (right + left)/ 2;
             if (countMissing(nums, m) >= k) {
-                r = m;
+                right = m;
+               
             } else {
-                l = m + 1;
+                left = m + 1;
             }
         }
 
-        return nums[l - 1] + (k - countMissing(nums, l - 1));
+        return nums[left - 1] + (k - countMissing(nums, left - 1));
     }
 
     private int countMissing(int[] nums, int i) {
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/medium/MissingElement.java
```diff
@@ -0,0 +1,35 @@
+package algorithms.curated170.medium;
+
+public class MissingElement {
```
altayhunoglu, line 3: Very good.

image

src/main/java/algorithms/curated170/medium/MissingElement.java
```diff
@@ -0,0 +1,35 @@
...
+            int m = l + (r - l) / 2;
+            if (countMissing(nums, m) >= k) {
+                r = m;
+            } else {
+                l = m + 1;
+            }
+        }
+
+        return nums[l - 1] + (k - countMissing(nums, l - 1));
+    }
```
altayhunoglu, line 19:
```suggestion
       public int missingElement(int[] nums, int k) {
        int left = 0;
        int right = nums.length;

        while (left < right) {
            int m = (right + left)/ 2;
            if (countMissing(nums, m) >= k) {
                right = m;
               
            } else {
                left = m + 1;
            }
        }

        return nums[left - 1] + (k - countMissing(nums, left - 1));
    }
```

src/main/java/algorithms/curated170/medium/MissingElementNaive.java
```diff
@@ -0,0 +1,28 @@
+package algorithms.curated170.medium;
+
+public class MissingElementNaive {
```
altayhunoglu, line 3: image

altayhunoglu (approved):
Valuable contribution. Thanks, Erdem.

PR #208, opened 2021-05-15
1533. Find the Index of the Large Integer

ErdemT09:
Issue: #39 
This solution utilizes a kind of binary search.
When we call the interface, what we are doing is comparing two ranges. The best way to do this is by comparing one half of the array with the another. Then, we are able to see which side of the array has the "Large Integer". We then move the corresponding range to that side of the array. For this, we do not need to keep track of all the 4 variables. Only knowing the upper and lower bounds of this range will be enough. These correspond to the variables l and y from the interface's method. r and x are determined by cutting this range in half and assigning respective values.  If the comparison yield 1, we move y to be r, as the sought value will be in that range of the array. Similarly, we move l to x if the comparison yields - 1, as the index will lie in that side of the array. This way, our range will eventually close in on the searched index.
If the comparison yields 0, the first time that this happens means that the 2 ranges have converged on each other enough to share a value, which will be the value `mid` in their first convergence.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/FindTheIndexOfTheLargeInteger.java
+++ b/src/main/java/algorithms/curated170/medium/FindTheIndexOfTheLargeInteger.java
@@ -4,7 +4,7 @@ public class FindTheIndexOfTheLargeInteger {
 
     public int getIndex(ArrayReader arr) {
         int length = arr.length();
-        int l = 0, y = length;
+      int l = 0, y = length-1;
         while (l < y) {
             int range = y - l + 1;
             int mid = (range - 1) / 2 + l;
--- a/src/main/java/algorithms/curated170/medium/FindTheIndexOfTheLargeInteger.java
+++ b/src/main/java/algorithms/curated170/medium/FindTheIndexOfTheLargeInteger.java
@@ -1,0 +1,33 @@
+package algorithms.curated170.medium;
+
+public class FindTheIndexOfTheLargeInteger {
+
+    public int getIndex(ArrayReader arr) {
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/FindTheIndexOfTheLargeInteger.java
```diff
@@ -0,0 +1,33 @@
+package algorithms.curated170.medium;
+
+public class FindTheIndexOfTheLargeInteger {
+
+    public int getIndex(ArrayReader arr) {
+        int length = arr.length();
+        int l = 0, y = length;
```
altayhunoglu, line 7:
```suggestion
      int l = 0, y = length-1;
```

altayhunoglu:
It works fine with 4ms~
I may suggest some readability fix. You can choose.

Especially, I avoid using lowercase "L" as a variable name as "1" and "l" looks so similar and makes it difficult to read.

```
class Solution {
  public int getIndex(ArrayReader arr) {
        int length = arr.length();
        int left = 0, right = length-1;

        while (left < right) {

            int half = (right - left +1) / 2;
    
            int c = arr.compareSub(left, left + half - 1, right - half + 1, right);
            
            if (c == 0) {
                return (right - left) / 2 + left;
            } else if (c < 0) {
                left = right - half + 1;
            } else {
                right = left + half - 1;
            }
        }
        return left;
    }
}
```

altayhunoglu (approved):
Thank you for your valuable collaboration Erdem.

PR #209, opened 2021-05-15
1229. Meeting Scheduler

ErdemT09:
Issue: #40 
The solution to this problem should be found as such:
1. We should find two intervals that intersect (If two intervals intersect, they can not belong to the same person according to constraints)
2. We should calculate the duration of this intersection
3. If that duration is longer than the sought duration, then return the beginning time of this interval and the beginning in addition with the sought duration.

There are two efficient approaches to this problem:

1) Sort & Compare:
We sort both given interval arrays according to their start times.
Then we create two pointers pointing to 2 arrays, beginning with the index 0.
Else, this implies that the intervals intersect. Interval's start time is the larger of two start times and lower of the two end times. If the difference between these lower and upper bounds is greater than the sought duration, then we return the List `{left, left + duration}`.  This can also be a negative number if intervals do not intersect, we don't return anything then.
If this is not the case, then, if the end time of the second interval is smaller than the first one, we increment the second pointer. Else, we increment the first pointer.
If we never return the above mentioned list, this means that no intervals sufficiently intersect we should return an empty list.

2) Priority Queue approach
We put all the intervals (that have a duration greater than the given duration) in a priority queue that compares these intervals according to their start times.
We then poll an element from this heap, and peek at the next. If the peeked interval's beginning after duration amount of time yields a time value that is smaller than the end value of the polled interval, then we return an interval beginning at the peeked interval and with duration.
If this never works, we return an empty list.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
+++ b/src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
@@ -1,23 +1,14 @@
 package algorithms.curated170.medium.meetingscheduler;
 
-import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 
 class MeetingScheduler {
     List<Integer> minAvailableDuration(int[][] p1, int[][] p2, int duration) {
 
-        Comparator<int[]> c = new Comparator<int[]>() {
-            @Override
-            public int compare(int[] o1, int[] o2) {
-                return o1[0] - o2[0];
-            }
-        };
-
-        Arrays.sort(p1, c);
-        Arrays.sort(p2, c);
+        Arrays.sort(p1, ((a, b) -> a[0] - b[0]));
+        Arrays.sort(p2, ((a, b) -> a[0] - b[0]));
 
         int i1 = 0, i2 = 0;
         while (i1 < p1.length && i2 < p2.length) {
--- a/src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
+++ b/src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
@@ -13,13 +13,7 @@ class MeetingScheduler {
         int i1 = 0, i2 = 0;
         while (i1 < p1.length && i2 < p2.length) {
             int s1 = p1[i1][0], e1 = p1[i1][1], s2 = p2[i2][0], e2 = p2[i2][1]; 
-
-            if (s2 >= e1) {
-                i1++;
-            } else if (s1 >= e2) {
-                i2++;
-                continue;
-            }
+            
             int left = Math.max(s1, s2); // lower bound of common interval
             int right = Math.min(e1, e2); // upper bound of common interval
             if (right - left >= duration) {
--- a/src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
+++ b/src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
@@ -12,7 +12,7 @@ class MeetingScheduler {
 
         int i1 = 0, i2 = 0;
         while (i1 < p1.length && i2 < p2.length) {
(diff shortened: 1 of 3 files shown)


src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
```diff
@@ -0,0 +1,40 @@
...
+
+class MeetingScheduler {
+    List<Integer> minAvailableDuration(int[][] p1, int[][] p2, int duration) {
+
+        Arrays.sort(p1, ((a, b) -> a[0] - b[0]));
+        Arrays.sort(p2, ((a, b) -> a[0] - b[0]));
+
+        int i1 = 0, i2 = 0;
+        while (i1 < p1.length && i2 < p2.length) {
+            int s1 = p2[i1][0], e1 = p2[i1][1], s2 = p2[i2][0], e2 = p2[i2][1];
```
altayhunoglu, line 15:
```suggestion
            int s1 = p1[i1][0], e1 = p1[i1][1], s2 = p2[i2][0], e2 = p2[i2][1]; 
```

ErdemT09: Thanks for noticing it.

src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingScheduler.java
```diff
@@ -0,0 +1,40 @@
...
+        int i1 = 0, i2 = 0;
+        while (i1 < p1.length && i2 < p2.length) {
+            int s1 = p2[i1][0], e1 = p2[i1][1], s2 = p2[i2][0], e2 = p2[i2][1]; 
+
+            if (s2 >= e1) {
+                i1++;
+            } else if (s1 >= e2) {
+                i2++;
+                continue;
+            }
```
altayhunoglu, line 22: I think you add this to prevent unnecessary max-min calculation.
 As it does not affect the efficiency, it is up to you removing or not.

altayhunoglu: image

ErdemT09: The unnecessary calculations done otherwise aren't that worse than calculating these. I will remove these then, for the sake of briefness.

altayhunoglu: > The unnecessary calculations done otherwise aren't that worse than calculating these. I will remove these then, for the sake of briefness.

You choose the right thing to do as expected. 😊

src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingSchedulerPriorityQueue.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Queue;
+
+class MeetingSchedulerPriorityQueue {
+    Queue<int[]> intervals;
+
+    List<Integer> minAvailableDuration(int[][] p1, int[][] p2, int duration) {
+
+        createHeap(p1, p2, duration);
+
+        while (!intervals.isEmpty()) {
```
altayhunoglu, line 15:
```suggestion
         while (intervals.size() > 1) {  //   When we poll from the queue, there should be any possibility that it is empty.
```

src/main/java/algorithms/curated170/medium/meetingscheduler/MeetingSchedulerPriorityQueue.java
```diff
@@ -0,0 +1,44 @@
+package algorithms.curated170.medium.meetingscheduler;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.PriorityQueue;
+import java.util.Queue;
+
+class MeetingSchedulerPriorityQueue {
```
altayhunoglu, line 8: image

altayhunoglu (approved):
Valuable collaboration.

PR #245, opened 2021-05-16
807. Max Increase to Keep City Skyline

ErdemT09:
[The question on LeetCode](https://leetcode.com/problems/max-increase-to-keep-city-skyline/submissions/)
This is a none-curated170 question that I have come across. When I noticed the solution idea, I wanted to document it here as well.
The question wants us to increase the numbers in the grid (i.e. the heights of the buildings), so that the skyline of the city from left-right or top-down does not change. The skyline is equivalent to the maximum height or the number in a column or a row. The fact that we need to increase the number without changing the skyline means that we have to increase the numbers without them changing the maximum value in their column and row. This means that numbers should be increased to the minimum of the maximum values of their respective columns and rows. This the intuition behind this solution.

LeetCode Submission result:
image

### DIFF
--- a/src/main/java/algorithms/medium/MaxIncreaseToKeepCitySkyline.java
+++ b/src/main/java/algorithms/medium/MaxIncreaseToKeepCitySkyline.java
@@ -2,7 +2,7 @@ package algorithms.medium;
 
 public class MaxIncreaseToKeepCitySkyline {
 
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Thanks. Nice question and a clean solution.

PR #246, opened 2021-05-16
245. Shortest Word Distance III

ErdemT09:
Issue: #147 
This solution is just a slight modification of [243. Shortest Word Distance ](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/easy/ShortestWordDistance.java). The only difference is that, if the word at i1 the same same as the word at the current index, shifting i1 to index i would have us lose information about the location of the current index, which would make the question unsolvable if two given words are the same. 
To fix that, i1 must only be shifted, if i1 is not assigned (equals to -1) or if the word at this previous index is not equal to i1.

There is also an alternative solution like this that makes further use of the fact that both words can be the same: https://cheonhyangzhang.gitbooks.io/leetcode-solutions/content/245-shortest-word-distance-iii-medium.html
I am not sure which solution is more elegant, that is something left to us to decide.


altayhunoglu:
It is the current fastest solution with 1ms.
But, I think you will not that much if-statements :)
```
class Solution {
    public int shortestWordDistance(String[] wordsDict, String word1, String word2) {
        int min = Integer.MAX_VALUE;
        if(word1.equals(word2)) {
            int prevIndex = -1;
            for(int i = 0; i < wordsDict.length; i++) {
                if(wordsDict[i].equals(word1)) {
                    if(prevIndex != -1) {
                        min = Math.min(min, i - prevIndex);
                    }
                    prevIndex = i;
                }
            }
        } else {
            int w1Index = -1;
            int w2Index = -1;
            for(int i = 0; i < wordsDict.length; i++) {
                if(wordsDict[i].equals(word1)) {
                    if(w2Index != -1) {
                        min = Math.min(min, i - w2Index);
                    }
                    w1Index = i;
                } else if(wordsDict[i].equals(word2)) {
                    if(w1Index != -1) {
                        min = Math.min(min, i - w1Index);
                    }
                    w2Index = i;
                }
            }
        }
        return min;
    }
}
```

ErdemT09:
It shouldn't be any faster, it most likely does the same amount of computations as the one with a single pointer.

ErdemT09:
```
public int shortestWordDistance(String[] words, String word1, String word2) {
        int idx = -1, res = Integer.MAX_VALUE;
        boolean same = word1 == word2;
        for (int i = 0; i < words.length; ++i) {
            if (words[i].equals(word1) || words[i].equals(word2)) {
                if (idx != -1 && (same || !words[i].equals(words[idx]))) {
                    res = Math.min(res, i - idx);
                }
                idx = i;
            }
        }
        return res;
    }
```
Doing both things together in abstracted/summarized way should be better.

altayhunoglu (approved):
Thanks for your valuable contribution.

ErdemT09:
I will mark this question in my mind as an example of how code can be shortened when logical conditions converge.

PR #248, opened 2021-05-16
Adds MovingAverageArray version, and moves the older one to the package.

altayhunoglu:
https://github.com/spiralgo/algorithms/issues/138
The array version takes 39ms. The older takes 44ms.


ErdemT09:
The other in is called array but the older solution is also an array solution.

altayhunoglu:
> The other in is called array but the older solution is also an array solution. And it is of the same time complexity as the other one. I think it's better to call it MovingAverage and our older solution MovingAverageAlternative.

Oke, it is done. Thanks.

ErdemT09 (approved):

ErdemT09:
The `buffer[]` is actually the data structure known as "circular buffer". 
https://en.wikipedia.org/wiki/Circular_buffer

altay9:
For the solution inside **MovingAverageAlternative**:

Here, we need to understand why, in the first place, we used "head" and "tail" pointers.

(Note: I think the naming of these variables can be confusing because of the nature of the buffer itself.
Because here, the array acts like a circular buffer. So, sometimes the tail can be the point on the left, sometimes it can be on the right, and vice versa.
### 1: `Head` is the pointer that will be pointing to the value that will be dequeued or removed from the array.
  
Why do we need to remove something from the array?
   
Because the array is fixed-sized, we want to have the most efficient use of space and time. 
Otherwise, we could use a queue instead. 
      
Which element should be overridden when more space is needed?
As we want our array to act like a queue, it will use FIFO, first-in, first-out.
So, the head will be pointing to the oldest element in the array.
When do we need to override an element?
We will have a fixed size of window of data we will read. We can assume the array itself is the window. So, we will need to use the places of oldest elements.

### 2-`Tail` is the pointer that will be pointing to the new element in the next method.

The trick to imitating a circular buffer is the modular arithmetic:
Head and tail increase constantly. But we are always looking at the indexes like that:
` head % window_size`  or `tail % window_size `

PR #249, opened 2021-05-16
Adds Logger Rate Limiter's fastest solution.

altayhunoglu:
https://github.com/spiralgo/algorithms/issues/140

image


ErdemT09:
https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/easy/Logger.java
What about this old code? It is basically the less brief version of this. We should delete it, I think.

altayhunoglu:
> https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/easy/Logger.java
> What about this old code? It is basically the less brief version of this. We should delete it, I think.

You are absolutely right. Done.

ErdemT09 (approved):
Simple and efficient code.

PR #250, opened 2021-05-16
The faster version takes 0ms, the older takes 3ms.

altayhunoglu:
https://github.com/spiralgo/algorithms/issues/144

image


ErdemT09:
Simplified:
```
public int largestUniqueNumber(int[] A) {
        int has [] = new int[1001];
        for(int i=0;i<A.length;i++){
            has[A[i]]+=1;
        }
        for(int i=1000;i>=0;i--){
            if(has[i]==1){
                return i;
            }
        }
        return -1; 
    }
```

ErdemT09:
The HashMap one can also be made more efficient:
//Will add it here

ErdemT09:
We might as well add something like this to the project while we are at it:
https://stackoverflow.com/a/1098137

PR #251, opened 2021-05-16
Palindrome permutation

altayhunoglu:
https://github.com/spiralgo/algorithms/issues/146
image

### DIFF
--- a/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumber.java
+++ b/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumber.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.easy;
+package algorithms.curated170.easy.largestuniquenumber;
  
 import java.util.HashMap;
 import java.util.Map;
--- a/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumberFaster.java
+++ b/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumberFaster.java
@@ -1,0 +1,19 @@
+package algorithms.curated170.easy.largestuniquenumber;
+
+ 
+public class LargestUniqueNumberFaster {
+  public int largestUniqueNumber(int[] A) {
+        int has [] = new int[1001];
+        for(int i=0;i<A.length;i++){
+            has[A[i]]+=1;
+        }
+        int res=-1;
+        for(int i=1000;i>=0;i--){
+            if(has[i]==1){
+                res=i;
+                break; 
+            }
+        }
+        return res; 
+    }
+}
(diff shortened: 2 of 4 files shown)


ErdemT09:
Instead of just calling it "-Faster", this time, it's better to call it "-Array" and the previous solution "-HashMap".

ErdemT09:
Uising 3 valued Java-logic might also be helpful here:
//Adding it.

altayhunoglu:
> Instead of just calling it "-Faster", this time, it's better to call it "-Array" and the previous solution "-HashMap".

Ok, this is done.

ErdemT09:
> Uising 3 valued Java-logic ( Optional<Boolean> )might also be helpful here:
> EDIT: We can just do it with normal booleans:
> 
> ```
> boolean sawSingle = false;
>             for (int key = 0; key < map.length; key++) {
>                 if (map[key] % 2 == 1) {
>                     if (sawSingle) {
>                         return false;
>                     }
>                     sawSingle = true;
>                 }
>             }
>             return true;
>         }
> ```

What do you think about the idea here? Would it work?

altayhunoglu:
> > Uising 3 valued Java-logic ( Optional )might also be helpful here:
> > EDIT: We can just do it with normal booleans:
> > ```
> > boolean sawSingle = false;
> >             for (int key = 0; key < map.length; key++) {
> >                 if (map[key] % 2 == 1) {
> >                     if (sawSingle) {
> >                         return false;
> >                     }
> >                     sawSingle = true;
> >                 }
> >             }
> >             return true;
> >         }
> > ```
> 
> What do you think about the idea here? Would it work?

Yeah, of course. Nice idea:

image


Even if, Optional class is not that useful in this context, It would be something like that:
 ```

            for (int key = 0; key < map.length; key++) {
                if (map[key] % 2 == 1) {
                   if (anExternalBooleanParameter.orElse(false)) {
                            return false;
                       }
                   anExternalBooleanParameter = Optional.of(true);
                }
            }
            return true;
 ```
 
 In what kind of context would an Optional be meaningful?
 In some cases like; reading the value of **anExternalBooleanParameter** (or another boolean) from a database or getting it via a REST service; we might not know if it is empty.
  For example, It would have been created like this and sent our API method as a parameter in an external service:

 ` Optional<Boolean> anExternalBooleanParameter= Optional.empty();`

  In that case, we should have given a default value using **orElse**.

ErdemT09:
Shall we add the boolean approach to the solution then?

altayhunoglu:
> Shall we add the boolean approach to the solution then?

Of course. I will add it too.

altayhunoglu:
> Shall we add the boolean approach to the solution then?

Oke, it is done.

ErdemT09:
Approved then.

ErdemT09 (approved):

PR #252, opened 2021-05-16
Max stack

altayhunoglu:
Issue: https://github.com/spiralgo/algorithms/issues/157

image


ErdemT09:
Again, we can come up with more clear namings: Such as 
For the first: MaxStack (this can actually remain so), MaxStackTwoStacks
For the second: MaxStackNode, MaxStackLinkedList, MaxStackLinkedNode
If there are any more explanatory names, we can switch to those too.

altayhunoglu:
> Again, we can come up with more clear namings: Such as
> For the first: MaxStack (this can actually remain so), MaxStackTwoStacks
> For the second: MaxStackNode, MaxStackLinkedList, MaxStackLinkedNode
> If there are any more explanatory names, we can switch to those too.

Oke, it is done. Thanks.

ErdemT09 (approved):

PR #253, opened 2021-05-16
Sentence similarity

altayhunoglu:
Issue: https://github.com/spiralgo/algorithms/issues/158

image

### DIFF
--- a/src/main/java/algorithms/curated170/easy/sentencesimilarity/SentenceSimilarity.java
+++ b/src/main/java/algorithms/curated170/easy/sentencesimilarity/SentenceSimilarity.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.easy;
+package algorithms.curated170.easy.sentencesimilarity;
  
 import java.util.HashSet;
 import java.util.List;
--- a/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumber.java
+++ b/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumber.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.easy;
+package algorithms.curated170.easy.largestuniquenumber;
  
 import java.util.HashMap;
 import java.util.Map;
--- a/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumberFaster.java
+++ b/src/main/java/algorithms/curated170/easy/largestuniquenumber/LargestUniqueNumberFaster.java
@@ -1,0 +1,19 @@
+package algorithms.curated170.easy.largestuniquenumber;
+
+ 
+public class LargestUniqueNumberFaster {
+  public int largestUniqueNumber(int[] A) {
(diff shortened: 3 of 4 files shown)


ErdemT09:
Why is the other one faster, it is after all a quadratic time algorithm whereas ours is O( n * log(n) ) (Considering that HashSet uses binary trees in the worst case)?

altayhunoglu:
> Why is the other one faster, it is after all a quadratic time algorithm whereas ours is O( n * log(n) ) (Considering that HashSet uses binary trees in the worst case)?

It is most probably about the difference between **contains** and **equals**.
If these words are not equal, then "equals" has O(1) time complexity. Contains method is always bigger than that.

ErdemT09:
But we loop through each string in the array and loop through the given word pairs each time. That is pseudo-quadratic complexity.  Our solution with the HashSet should be faster because we loop through each string and checking if an element is in the set takes O(log n) time at worst case, O(1) in average.

altayhunoglu:
> But we loop through each string in the array and loop through the given word pairs each time. That is pseudo-quadratic complexity.  Our solution with the HashSet should be faster because we loop through each string and checking if an element is in the set takes O(log n) time at worst case, O(1) in average. 


But to that, our older solution should do String concatenation which is also quite expensive. 
(As in Java String is immutable and it should create a copy of the string in each concatenation. It is why we needed StringBuilder.)

Also, as you mentioned the size of the array is a constant.

ErdemT09:
Instead of String concatenation, we can just use a list of two strings.

altayhunoglu:
> Instead of String concatenation, we can just use a list of two strings.

Oke, I will try with list too.

altayhunoglu:
> Instead of String concatenation, we can just use the given list pair of two strings:
> 
> ```
> public boolean areSentencesSimilar(
>             String[] words1, String[] words2,List<List<String>> pairs) {
>         if (words1.length != words2.length) return false;
> 
>         Set<List<String>> pairset = new HashSet<>();
>         for (List<String> pair: pairs)
>             pairset.add(pair);
> 
>         for (int i = 0; i < words1.length; ++i) {
>             if (!words1[i].equals(words2[i]) &&
>                     !pairset.contains(List.of(words1[i], words2[i])) &&
>                     !pairset.contains(List.of(words2[i], words1[i])))
>                 return false;
>         }
>         return true;
>     }
> ```

Oke, it is the result:
image

Even List.of has its own price, but it is better than the older version (as the older one takes 9ms)

We can add those three solutions.

src/main/java/algorithms/curated170/easy/sentencesimilarity/SentenceSimilarity.java
```diff
@@ -21,4 +21,20 @@ public boolean areSentencesSimilar(
         }
         return true;
     }
+  public boolean areSentencesSimilarWithoutConcatination(
```
altayhunoglu, line 24: Adds the solution without concatenation that takes 1ms instead of 9ms.

ErdemT09 (approved):

Issue #262, opened 2021-05-17
723. Candy Crush


altayhunoglu:
https://leetcode.com/problems/candy-crush

Varsayılan kod:
```
class Solution {
    public int[][] candyCrush(int[][] board) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118418520-67b83180-b6c1-11eb-810b-80eebcdc6f8f.png)
![image](https://user-images.githubusercontent.com/60903744/118418528-6e46a900-b6c1-11eb-99fb-c390b210ec47.png)

**Explanation:**

![candy_crush_example_2](https://user-images.githubusercontent.com/60903744/118418547-7dc5f200-b6c1-11eb-901f-69b7fa4e87bc.png)
![image](https://user-images.githubusercontent.com/60903744/118418560-861e2d00-b6c1-11eb-850a-4b5f35ac4089.png)

altayhunoglu:
Solved: https://github.com/spiralgo/algorithms/pull/263

PR #263, opened 2021-05-17
723. Candy Crush

altayhunoglu:
https://github.com/spiralgo/algorithms/issues/262

image


src/main/java/algorithms/curated170/medium/candycrush/CandyCrushFaster.java
```diff
@@ -0,0 +1,68 @@
...
+                            target[row][col] = 0;
+                        }
+                        targetIndex--;
+                    }
+                }
+            }
+        }
+
+        return crush;
+    }
```
ErdemT09, line 60: We can extract this to a method.
```suggestion
if (crush) {
            drop(target);
        }
        return crush;
    }

    private void drop(int[][] target) {
        for (int col = 0; col < target[0].length; col++) {
            int targetIndex = target.length - 1;
            for (int row = target.length - 1; row >= 0; row--) {
                if (target[row][col] != 0) {
                    if (row != targetIndex) {
                        target[targetIndex][col] = target[row][col];
                        target[row][col] = 0;
                    }
                    targetIndex--;
                }
            }
        }
    }
```

ErdemT09:
I don't think we need the old solution besides its use of recursion. 
Here's the new solution with recursion used:
```
public int[][] candyCrush(int[][] board) {
        int[][] target = new int[board.length][board[0].length];
        if(crush(board, target))
        {
            return candyCrush(target);
        }
        return board;
    }
```

altayhunoglu:
> I don't think we need the old solution besides its use of recursion. 
> 
> Here's the new solution with recursion used:
> 
> ```
> 
> public int[][] candyCrush(int[][] board) {
> 
>         int[][] target = new int[board.length][board[0].length];
> 
>         if(crush(board, target))
> 
>         {
> 
>             return candyCrush(target);
> 
>         }
> 
>         return board;
> 
>     }
> 
> ```

We can simply remove the older.

altayhunoglu:
Oke. We have applied your modularity suggestion.

altayhunoglu:
> Don't we delete the older solution and apply its recursiveness to the newer one?

I am sorry that I recognized you emphasized on the recursive part yet.

Sure, I will try to do it.

altayhunoglu:
> Don't we delete the older solution and apply its recursiveness to the newer one?

Oke, it is done.

src/main/java/algorithms/curated170/medium/candycrush/CandyCrush.java
```diff
@@ -12,132 +10,79 @@ public static void main(String[] args) {
...
+    public int[][] candyCrush(int[][] board) {
+        int[][] target = new int[board.length][board[0].length];
 
-                            boardCopy[i][low] = board[i][low];
-                            low++;
+        if (crush(board, target)) {
+            int[][] temp = board;
+            board = target;
+            target = temp;
+            return candyCrush(board);
```
ErdemT09, line 27: Why do we assign  values to temp and target here when we are not using them again?

ErdemT09: We can just input target to the method.

altayhunoglu:
You are right. It is just a mistake.

ErdemT09:
Smart refactoring for the last one.
For this question, it kind of reminded me the concept of "game loop" from game development:
image

ErdemT09 (approved):

altayhunoglu:
Thanks for the information and attention Erdem.

Issue #264, opened 2021-05-17
1100. Find K-Length Substrings With No Repeated Characters


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int numKLenSubstrNoRepeats(String s, int k) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118418762-78b57280-b6c2-11eb-9c82-e5b7d413a9fb.png)
![image](https://user-images.githubusercontent.com/60903744/118418766-7ce19000-b6c2-11eb-82ec-310fa7e0cfdc.png)

altayhunoglu:
Solved: https://github.com/spiralgo/algorithms/pull/265

PR #265, opened 2021-05-17
1100. Find K-Length Substrings With No Repeated Characters

altayhunoglu:
Issue:

https://github.com/spiralgo/algorithms/issues/264

image


ErdemT09:
Again, calling one solution HashSet and the other Array is more readable.

altayhunoglu:
> Again, calling one solution HashSet and the other Array is more readable.

Thank you, it is done.

ErdemT09 (approved):

Issue #271, opened 2021-05-18
750. Number Of Corner Rectangles


altayhunoglu:
Varsayılan kod:
```
class Solution {
    public int countCornerRectangles(int[][] grid) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118682960-9b9e6e80-b809-11eb-8f55-e80663cc6704.png)
![image](https://user-images.githubusercontent.com/60903744/118682986-a35e1300-b809-11eb-9716-d57f5054109f.png)
![image](https://user-images.githubusercontent.com/60903744/118683007-a822c700-b809-11eb-88f0-1384a38b083c.png)

altayhunoglu:
solved: https://github.com/spiralgo/algorithms/pull/272

PR #272, opened 2021-05-18
750. Number Of Corner Rectangles

altayhunoglu:
Issue: https://github.com/spiralgo/algorithms/issues/271
The older solution did not work. I add the current fastest one with 10 ms.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/NumberOfCornerRectangles.java
+++ b/src/main/java/algorithms/curated170/medium/NumberOfCornerRectangles.java
@@ -15,9 +15,8 @@ public class NumberOfCornerRectangles {
     
    public int countCornerRectangles(int[][] grid) {
         int totalCount = 0;
-        int rows = grid.length;
         int cols = grid[0].length;
-        int[][] counts = new int[rows][cols];
+        int[][] counts = new int[cols][cols];
         for (int[] row : grid) {
             for (int col1 = 0; col1 < cols - 1; col1++) {
                 if (row[col1] == 1) {
--- a/src/main/java/algorithms/curated170/medium/NumberOfCornerRectangles.java
+++ b/src/main/java/algorithms/curated170/medium/NumberOfCornerRectangles.java
@@ -32,4 +32,19 @@ public class NumberOfCornerRectangles {
         }
         return totalCount;
     }
+ 
+ public int  countCornerRectanglesRowFirstApproach(int[][] grid) {
+        int totalCount= 0;
+        for (int i = 0; i < grid.length - 1; i++) {
+            for (int j = i + 1; j < grid.length; j++) {
+                int count = 0;
+                for (int k = 0; k < grid[0].length; k++) {
+                    if (grid[i][k] == 1 && grid[j][k] == 1) count++;
+                }
+                if (count > 0) totalCount+= count * (count - 1) / 2;
+            }
+        }
+        return totalCount;
+    }
+ 
 }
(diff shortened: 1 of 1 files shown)


ErdemT09:
I will write an explanation here when I understand:

ErdemT09:
I have actually came across this question while I was watching the mock Google interview by the guy from AlgoExpert with this competitive programmer.  https://youtu.be/EuPSibuIKIg (Watching the video itself is not useful though.)
This question might be an important coding question then.

src/main/java/algorithms/curated170/medium/NumberOfCornerRectangles.java
```diff
@@ -13,21 +13,24 @@ public static void main(String[] args) {
...
-      
-      }
-      return count;
-      
- }
+   public int countCornerRectangles(int[][] grid) {
+        int totalCount = 0;
+        int rows = grid.length;
+        int cols = grid[0].length;
+        int[][] counts = new int[rows][cols];
```
ErdemT09, line 20: Why don't we set the size to int[cols][cols]? Don't we only assign values to cols here?

ErdemT09:
There is also this enumarating solution.
```
 public int _countCornerRectangles(int[][] grid) {
        int ans = 0;
        for (int i = 0; i < grid.length - 1; i++) {
            for (int j = i + 1; j < grid.length; j++) {
                int counter = 0;
                for (int k = 0; k < grid[0].length; k++) {
                    if (grid[i][k] == 1 && grid[j][k] == 1) counter++;
                }
                if (counter > 0) ans += counter * (counter - 1) / 2;
            }
        }
        return ans;
    }
```
Basically, we compare every single row with each other and add the rectangles that we find. This has a time complexity of `O(  row^2 * col )`, whereas the current solution has a complexity of `O(  col^2 * row )`.

src/main/java/algorithms/curated170/medium/NumberOfCornerRectangles.java
altayhunoglu: Yep, you are absolutely right. Somehow, it passes all LeetCode test cases. 
It means test cases are not rigorous enough.
Thanks for your attention. We fix it.

altayhunoglu:
> There is also this enumarating solution.
> 
> ```
>  public int _countCornerRectangles(int[][] grid) {
>         int totalCount= 0;
>         for (int i = 0; i < grid.length - 1; i++) {
>             for (int j = i + 1; j < grid.length; j++) {
>                 int count = 0;
>                 for (int k = 0; k < grid[0].length; k++) {
>                     if (grid[i][k] == 1 && grid[j][k] == 1) count++;
>                 }
>                 if (count > 0) totalCount+= count * (count - 1) / 2;
>             }
>         }
>         return totalCount;
>     }
> ```
> 
> Basically, we compare every single row with each other and add the rectangles that we find. This has a time complexity of `O( row^2 * col )`, whereas the current solution has a complexity of `O( col^2 * row )`.

image

ErdemT09:
Should we add the slower row-first based approach too?

altayhunoglu:
> Should we add the slower row-first based approach too?

If you think it is necessary, of course we should add.

altayhunoglu:
> > > Should we add the slower row-first based approach too?
> > 
> > 
> > If you think it is necessary, of course we should add.
> 
> I think we should, it has its own special sides as it is summing all numbers up to a certain count. Besides, it has a spatial complexity of O ( 1 ) and doesn't exchange that much speed for that.

Oke, it is done.

ErdemT09 (approved):

PR #274, opened 2021-05-18
Created the first Wiki Page

altayhunoglu:
I will migrate all the questions to Wiki.


ErdemT09 (approved):
A well start.
Small problem:
The quality of the image in the Wiki page is actually low, we should replace it with a higher quality version of the image.

altayhunoglu:
> A well start.
> 
> Small problem:
> 
> The quality of the image in the Wiki page is actually low, we should replace it with a higher quality version of the image.

You are absolutely right.
I will do it when I am available.

PR #275, opened 2021-05-19
1066. Campus Bikes II

ErdemT09:
Note: I tried recording a video for this but things didn't work out, so I will try to explain it with some text and drawings:

Let's say we have 4 workers, indexed as in the question: `0 1 2 3` 
And 4 bikes, indexed again as such: `0 1 2 3` 
We can represent the state of which bikes have been picked using a binary representation `1`'s or `0`'s or with boolean values of `true` and `false`.
Binary representation is the one used in this question. For example, our initial 4 unpicked bikes will have the binary representation of `1111`. In this example we need to find the minimum sum of distances for this given state of bikes.
To solve this problem, we need to solve sub-problems consisting of making decision about which worker should get which bike.
First, we begin by assigning the bikes to our worker at index `0` and increase the index of our worker for consistency.
This gives rise to the following decision pattern:
image
After assigning a bike to the `0`th and the `1`st worker, we have to also have to make a decision about how the remaining two bikes should be assigned to worker `2` and `3`. But in the decision tree, it is visible that we are going to make the same calculations again. Instead of calculating them again, we can store the minimum distance result of that state in an array, so that we do not have to calculate that state again: 
image
Here are few states that reappear again. The careful reader can notice even more. 
With that, we can solve this question in order of O(2 ^ b) time, as we only have to make decisions for each state of bikes. The actual time complexity will however be something like O ( w *  b * 2 ^ b), because for each worker, we loop through each bike, and resolving a decision about every bike at that state.

The binary operations required for this algorithm are quite simple. To obtain an integer with binary representation of such, we just have to shift 1 b bits to the right and subtract 1 from it:
As for the example: 
Shift 4 bits: `10000` 
Substract 1: `10000-1 = 1111`
We can also turn off bits to signal taken bikes using the xor operator.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/CampusBikes2.java
+++ b/src/main/java/algorithms/curated170/medium/CampusBikes2.java
@@ -12,10 +12,10 @@ public class CampusBikes2 {
 
       private int targetBitmask;
       int[][] workers, bikes;
-      int[][] dp;
+      int[] dp;
 
       public int assignBikes(int[][] workers, int[][] bikes) {
-            dp = new int[workers.length][1<<bikes.length];
+            dp = new int[1<<bikes.length];
             
             this.workers = workers;
             this.bikes = bikes;
@@ -28,9 +28,9 @@ public class CampusBikes2 {
             {
                   return 0;
             }
-            if(dp[worker][mask] != 0)
+            if(dp[mask] != 0)
             {
-                  return dp[worker][mask];
+                  return dp[mask];
             }
             int ans = Integer.MAX_VALUE;
             for(int j = 0; j<bikes.length; j++)
@@ -40,13 +40,13 @@ public class CampusBikes2 {
                         ans = Math.min(ans, dist(worker, j) + DFS(worker+1, mask^(1<<j)));
                   }
             }
-            return dp[worker][mask] = ans;
+            return dp[mask] = ans;
       }
       public static void main(String[] args) {
 
             var solution = new CampusBikes2();
-            int[][] workers = leetCodeIntegerGridConverter("[[0,0],[1,1],[2,0],[3,4]]");
-            int[][] bikes = leetCodeIntegerGridConverter("[[1,0],[2,2],[2,3],[4,0]]");
+            int[][] workers = leetCodeIntegerGridConverter("[[1,1],[1,3],[1,5],[4,5]]");
+            int[][] bikes = leetCodeIntegerGridConverter("[[4,2],[2,2],[3,5],[5,1]]");
 
             int sol = solution.assignBikes(workers, bikes);
             System.out.println(sol);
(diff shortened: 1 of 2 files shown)


ErdemT09:
I have also added a array converter for converting Python/LeetCode/Java arrays in console print to actual 2 dimensional Java arrays. What should we do with it?

ErdemT09:
I have also added a array converter for converting Python/LeetCode/Java arrays in console print to actual 2 dimensional Java arrays. What should we do with it?

altayhunoglu:
> I have also added a array converter for converting Python/LeetCode/Java arrays in console print to actual 2 dimensional Java arrays. What should we do with it?

We can add into repo just like a usual class file.

src/main/java/algorithms/curated170/medium/CampusBikes2.java
```diff
@@ -0,0 +1,89 @@
...
+
+            var solution = new CampusBikes2();
+            int[][] workers = leetCodeIntegerGridConverter("[[0,0],[1,1],[2,0],[3,4]]");
+            int[][] bikes = leetCodeIntegerGridConverter("[[1,0],[2,2],[2,3],[4,0]]");
+
+            int sol = solution.assignBikes(workers, bikes);
+            System.out.println(sol);
+      }
+
+      public static int[][] leetCodeIntegerGridConverter(String arr) {
```
altayhunoglu, line 55: You can create a package named "utils" or "util" and move this under a custom class.
You can name the class something like LeetCodeUtil as we can maintain and enhance it in the future.

src/main/java/algorithms/curated170/medium/CampusBikes2.java
```diff
@@ -0,0 +1,89 @@
...
+      private int dist(int wPos, int bPos) {
+            return Math.abs(workers[wPos][0] - bikes[bPos][0]) + Math.abs(workers[wPos][1] - bikes[bPos][1]);
+      }
+
+      private int targetBitmask;
+      int[][] workers, bikes;
+      int[][] dp;
+
+      public int assignBikes(int[][] workers, int[][] bikes) {
+            dp = new int[workers.length][1<<bikes.length];
```
altayhunoglu, line 18: It looks we may not really need which specific worker has the bike. If we use "mask" as a database to persist the state of bike assignment, dp can be 1D.

ErdemT09: True, this holds valid because only n many digits could have been 0'ed when 2 workers are assigned;

ErdemT09: Added.

altayhunoglu (approved):
Thank you Erdem. 
Your collaboration is so valuable as always. 
Approved.

ErdemT09:
> Thank you Erdem.
> Your collaboration is so valuable as always.
> Approved.

I have added an explanation now.
How does this algorithm perform on LeetCode?

altayhunoglu:
> > Thank you Erdem.
> > Your collaboration is so valuable as always.
> > Approved.
> 
> I have added an explanation now.
> How does this algorithm perform on LeetCode?

It is remarkably fast.
Thanks again for the explanation by the way. No worries about the video. You can do it whenever you want and I am sure you will nail it too.

image

ErdemT09:
Can I get an approve again?
I had to remove an unused variable that I forgot.

altayhunoglu (approved):

PR #280, opened 2021-05-20
358. Rearrange String k Distance Apart

ErdemT09:
Issue: #238 
Firstly, it is always better to have been finished the medium problems. However, when I read this question, this solution immediately occurred to me I wanted to document it.

# The Algorithm

## Approach 1 - Layers (with Priority Queue and Array)

 **Point 1:**
To write an algorithm, we need consistency with what we are doing. The consistency that comes to mind in such a question would be to arrange this string layer by layer. A basic way to do this is to place the characters of the same layer by their ASCII order.

 **Point 2:**
Always place the string with highest number of occurrences. For this, we have to count how many times a letter occurs in the input string `s` and place characters with the most counts the most prior. We also place each character in a layer not more than once. If we do place, that means that `s` can not be rearranged.
For example, consider the input `("abbc", 3)`. 
Here, just placing them alphabetically would have yielded the result `abcb`, which is false. The right way to rearrange this would have been: First placing the chars with the highest priorities, then try completing the layer, then checking if the layer is completed. 
This now yields the result: `bacb`, which is correct.
The priorities of characters get changed as they are placed, so our algorithm needs to adapt the needed dynamism for that.

 **Point 3:**
Expanding upon point 2, another thing that we should do is to stop when the layers are complete.
For example, consider the input `("aadbbbcce", 4)`. 
If we place the letters only considering their count without stopping, our algorithm would yield the result `"bacdebacb"`. In the first layer, after successfully placing 4 characters, we didn't stop and instead placed even more characters. Because of this, we did not have enough characters for the next layer and our algorithm resulted in a false string. 
In order not to waste layers, we must know when to stop, which is when we complete the layer of length `k`, here (4). This results in the string `"bacdbaceb"`, which is a correct result.

## Approach 2 - Find the current best and keep its index

First, we count how many times each letter occurs in the input string and store them in a count-database.
Here, similar to the approach above, we poll out letters from our letter database (An array, for example). Based on 2 criteria:
1. The letter we choose shouldn't have been encountered at least `k` digits ago.
2. It must have the highest count.
This means that we iterate through every 26 letter for every character that we place in the rearranged string. If no letter fulfills the first criteria, that means that we cannot rearrange the string.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/RearrangeStringKDistanceApart.java
+++ b/src/main/java/algorithms/curated170/hard/RearrangeStringKDistanceApart.java
@@ -1,58 +1,122 @@
 package algorithms.curated170.hard;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Set;
-import java.util.Map.Entry;
+import java.util.List;
+import java.util.PriorityQueue;
 
 public class RearrangeStringKDistanceApart {
-    private final int unassigned = -1073741824;
+    /**
+     *
+     */
+    private static final int ALPHABET_SIZE = 26;
+    private static final int UNASSIGNED = -1073741824;
+
+    int[] counter;
+    HashMap<Integer, PriorityQueue<Character>> layers;
+
     public String rearrangeString(String s, int k) {
 
-        HashMap<Character, Integer> counter = new HashMap<>();
-        for (char c : s.toCharArray()) {
-            counter.put(c, counter.getOrDefault(c, 0) + 1);
-        }
-        var keys = counter.keySet();
-        
-        StringBuilder rearrangedStr = new StringBuilder();
-        int[] lastIndices = new int[26];
-        Arrays.fill(lastIndices, unassigned);
-        int index = 0; 
-        
+        countLetters(s);
 
-        while (!counter.isEmpty()) {
-            for (char key : keys.toArray(new Character[0])) {
-                if (!counter.containsKey(key)) {
+        setLetterLayers();
+
+        StringBuilder rearrangedStr = new StringBuilder();
+
+        while (!layers.isEmpty()) {
+
+            List<Integer> keyList = Arrays.asList(layers.keySet().toArray(new Integer[0]));
+            Collections.reverse(keyList);
+
+            HashMap<Integer, PriorityQueue<Character>> newLayers = new HashMap<>();
+            int layerBreadth = k-1;
+            for (int layer : keyList) {
+                var pq = layers.get(layer);
+
+                if (layerBreadth < 0) {
+                    if (!newLayers.containsKey(layer)) {
+                        newLayers.put(layer, pq);
+                    } else {
+                        newLayers.get(layer).addAll(pq);
+                    }
                     continue;
                 }
-                rearrangedStr.append(key);
-                counter.put(key, counter.get(key) - 1);
-                if (counter.get(key) == 0) {
-                    counter.remove(key);
+
+                PriorityQueue<Character> nextLayer = new PriorityQueue<>();
+                while (!pq.isEmpty()) {
+                    char c = pq.poll();
+                    if (layerBreadth > -1) {
+                        rearrangedStr.append(c);
+                        nextLayer.add(c);
+                        layerBreadth--;
+                    } else {
+                        newLayers.putIfAbsent(layer, new PriorityQueue<>());
+                        newLayers.get(layer).add(c);
+                    }
                 }
-                if((index - lastIndices[letterIndex(key)]) < k)
+                if(layer > 1)
                 {
-                    return "";
+                    newLayers.put(layer - 1, nextLayer);
                 }
-                lastIndices[letterIndex(key)] = index;
-                index++;
             }
+
+            if(!newLayers.isEmpty() && layerBreadth > -1)
+            {
+                return "";
+            }
+            layers = newLayers;
         }
-        return rearrangedStr.toString();    
+
+        return rearrangedStr.toString();
     }
 
-    private final int LETTER_a = 97;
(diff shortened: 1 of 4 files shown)


ErdemT09:
I have also came up with this code, but this should not work as wel:

```
public class RearrangeStringKDistanceApart {
    private static final int ALPHABET_SIZE = 26;
    private static final int UNASSIGNED = -1073741824;

    public String rearrangeString(String s, int k) {

        int[] counter = new int[ALPHABET_SIZE];
        for (char c : s.toCharArray()) {
            counter[getIndexOfLetter(c)]++;
        }

        HashMap<Integer, PriorityQueue<Character>> layers = new HashMap<>();
        for (int i = 0; i < ALPHABET_SIZE; i++) {
            int count = counter[i];
            if (count == 0) {
                continue;
            }
            layers.putIfAbsent(count, new PriorityQueue<>());
            layers.get(count).add(getLetterOfIndex(i));
        }
        System.out.println(layers);

        StringBuilder rearrangedStr = new StringBuilder();
        int[] lastIndices = new int[ALPHABET_SIZE];
        Arrays.fill(lastIndices, UNASSIGNED);
        int index = 0;

        while (!layers.isEmpty()) {
            List<Integer> keyList = Arrays.asList(layers.keySet().toArray(new Integer[0]));
            Collections.reverse(keyList);
            HashMap<Integer, PriorityQueue<Character>> newLayers = new HashMap<>();
            for (int layer : keyList) {
                var pq = layers.get(layer);
                boolean hasNext = layer > 1;
                PriorityQueue<Character> nextLayer = new PriorityQueue<>();
                while (!pq.isEmpty()) {
                    char c = pq.poll();
                    rearrangedStr.append(c);
                    if (hasNext) {
                        nextLayer.add(c);
                    }
                }
                if (hasNext) {
                    newLayers.put(layer - 1, nextLayer);
                }
            }
            layers = newLayers;
        }
        System.out.println(rearrangedStr.toString());
        return rearrangedStr.toString();
    }

    private static final int LETTER_a = 97;

    private int getIndexOfLetter(char key) {
        return key - LETTER_a;
    }

    private char getLetterOfIndex(int index) {
        return (char) (LETTER_a + index);
    }

        var solution = new RearrangeStringKDistanceApart();
        solution.rearrangeString("aadbbbcce", 1); // prints bacdbacb
        solution.rearrangeString("bab", 1); // prints bab
    }
}

```

ErdemT09:
I have added a reversed iterator class that I have seen here https://stackoverflow.com/a/1098153.
I have also improved it a bit.
Here is an implementation of it:
```
import algorithms.util.collections.*;

public class Playground {

        Map<Character, Integer> map = new TreeMap<Character, Integer>() {
            {
                put('z', 2);
                put('t', 9);
                put('e', 7);
                put('q', 3);
                put('n', 5);
                put('a', 4);
            }
        };

        System.out.println("For sets: ");
        for (char a : Reversed.reversed((map.keySet())))
        {
            System.out.print(a + " ");
        }
        System.out.println();
        List<Integer> k = List.of(10,7,9,6,4,5,1,12,2,3);
        
        System.out.println("For lists: ");
        for (int a : Reversed.reversed(k))
        {
            System.out.print(a + " ");
        }
    }

}

Prints:
```
For sets: 
z t q n e a 
For lists: 
3 2 12 1 5 4 6 9 7 10 
```
```

altayhunoglu:
> I have added a reversed iterator class that I have seen here https://stackoverflow.com/a/1098153.
> I have also improved it a bit.

 Thanks for the consideration Erdem. Nice util.

ErdemT09:
stackedit.io
Before I write the explanation, I would like to note this site here as a helper for writing markdown texts in GitHub.

ErdemT09:
Another small note:
image

image

ErdemT09:
Explanation for the first approach added.

altayhunoglu:
Wonderfully explained Erdem. Thanks.

ErdemT09:
I still need to add an explanation to the other approach.

altayhunoglu (approved):
I think both of these solutions are fine enough.
 If we need, we can think about some ways to make the code simpler and leaner.

 By the way, just as a note for the future, I add the current fastest solution on LeetCode. No need to add it into our codebase as it is not readable and modular:
 
```java
 class Solution {
    public String rearrangeString(String s, int k) {
        if (k == 0) return s;
        int[] freq = new int[26];
        for (char c: s.toCharArray()) {
            freq[c-'a']++;
        }
        int max = 0;
        char maxchar = 'a';
        int maxNum = 0;
        for (int i = 0; i < freq.length; i++) {
            if (freq[i] > max) {
                max = freq[i];
                maxchar = (char) (i + 'a');
                maxNum = 1;
            } else if (freq[i] == max) {
                maxNum++;
            }
        }
        int part = max - 1;
        int slot = part * (k - maxNum);
        if ( slot > s.length() - max * maxNum) return "";
        char[] res = new char[s.length()];
        Arrays.fill(res, '_');
        int p = 0;
        while (freq[maxchar-'a']-- > 0) {
            res[p] = maxchar;
            p+=k;
        }
        for (int i = 0; i < freq.length; i++) {
            if (freq[i] <= 0) continue;
            // p = 0;
            char curr = (char) (i + 'a');
            while (freq[i]-- > 0) {
                if (p >= res.length) p = 0;
                while (res[p] != '_') {
                    p = (p + 1) % res.length;
                }
                res[p] = curr;
                p += k;
            }
            p = p-k+1;
        }
        return String.valueOf(res);
    }
}

```

ErdemT09:
I have also add the explanation for the faster, other approach. Which is a more understandable version of this code that we haven't added.

ErdemT09:
Note here: The second approach can also be achieved with a priority queue.

altayhunoglu:
> I have also add the explanation for the faster, other approach. Which is a more understandable version of this code that we haven't added.

Thank you Erdem.

PR #281, opened 2021-05-20
320. Generalized Abbreviation

ErdemT09:
Issue: #43 

## Algorithm
There are 2 approaches to this problem:

1. Bit Manipulation
2. Backtracking

I will begin with explaining the bitwise solution as I have found to be actually more straightforward:

### Approach-1 Bit Manipulation:
We can encode abbreviated strings in binary as such:
For example, let `s` be the string `"word"`
The abbreviation: 2rd, can be encoded in binary as 1100
The abbreviation: w1r1, can be encoded in binary as 0101
This way, we can iterate through all the possible permutations of 0's and 1's in the length of the given string, which is equal to iterating through all numbers up to 2^word.length, only the 0's in the searched part are considered. convert these permutations to strings by looping through the digits. If two or more 1's are next to each other, we naturally sum these as we did in our example.

### Approach-2 Branching and Backtracking:

At every index of the string, we have to make a decision about whether we abbreviate the string or keep the char. If we have made previous abbreviation decisions, we add them if we decide to keep the character. This creates two branches of methods in a recursive manner. Once these additions are complete, when the index reaches the end of the string and the methods return, we must reset the length of the string builder so that the upper branches of the current method and their children are not effected by the changes made to the string builder during the branching. This is how we "backtrack".

### DIFF
--- a/src/main/java/algorithms/curated170/medium/GeneralizedAbbreviation.java
+++ b/src/main/java/algorithms/curated170/medium/GeneralizedAbbreviation.java
@@ -1,0 +1,47 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class GeneralizedAbbreviation {
+    public List<String> generateAbbreviations(String word) {
+        List<String> ans = new ArrayList<>();
+
+        // the key is the binary representation of how many and which chars are converted to digits
+        //1100 would show that first two chars are the number 2 now
+        for (int key = 0; key < (1 << word.length()); key++)
+        {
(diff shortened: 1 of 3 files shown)


altayhunoglu (approved):
Both solutions are quite fine. 
I tried to draw the decision tree for the backtracking solution.
It uses some kind of DFS `in-order traverse`.

IMG-1628

ErdemT09:
> Both solutions are quite fine. 
> I tried to draw the decision tree for the backtracking solution.
> It uses some kind of DFS `in-order traverse`.
> 
> IMG-1628
> 

As we go further down the ree, it becomes more costly to rebuild each shared string. That is why the bitwise solution is way slower without any additional dynamic programming methods.

Issue #231, opened 2021-05-15
499. The Maze III


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380744-d37c9a80-b5ec-11eb-9892-ba262233fe54.png)
![maze3-1-grid](https://user-images.githubusercontent.com/60903744/118380747-d9727b80-b5ec-11eb-8ae3-62c322d4db41.jpg)
![image](https://user-images.githubusercontent.com/60903744/118380748-de372f80-b5ec-11eb-8c22-1a07526e1d29.png)
![maze3-2-grid](https://user-images.githubusercontent.com/60903744/118380751-e4c5a700-b5ec-11eb-90ae-67c2b44b804d.jpg)
![image](https://user-images.githubusercontent.com/60903744/118380755-ea22f180-b5ec-11eb-810a-a22ed290fca5.png)
![image](https://user-images.githubusercontent.com/60903744/118380759-ef803c00-b5ec-11eb-9ae0-255742ed8a1c.png)

Issue #56, opened 2021-03-08
505. The Maze II


altayhunoglu:
https://github.com/altayhunoglu/algorithms/issues/48 sorusu ile aynı.
Ama bu sefer, hedefe varmanın mümkün olup olmadığını değil, hedefe giden en kısa mesafeyi ölçmemizi istiyor.

LeetCode: https://leetcode.com/problems/the-maze-ii

Varsayılan kod:

```
class Solution {
    public int shortestDistance(int[][] maze, int[] start, int[] destination) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/117371058-f792fa00-aecf-11eb-94d3-720e31db7f4b.png)
![image](https://user-images.githubusercontent.com/60903744/117371100-05e11600-aed0-11eb-8533-91a90f995735.png)
![image](https://user-images.githubusercontent.com/60903744/117371137-0ed1e780-aed0-11eb-88b3-fc88df7709c7.png)
![image](https://user-images.githubusercontent.com/60903744/117371162-17c2b900-aed0-11eb-8761-ca7a95e94926.png)

PR #282, opened 2021-05-21
Adds a DFS and a BFS solution for "The Maze"

altayhunoglu:
Resolves https://github.com/spiralgo/algorithms/issues/48
 
 This solution is very important to grasp DFS and it is a kind of basis for [The Maze II](https://github.com/spiralgo/algorithms/issues/56) and [The Maze III](https://github.com/spiralgo/algorithms/issues/231).
 
 It is also a little bit different than the usual DFS problems as you cannot stop and change direction until you encounter a wall or the destination.
 
 It is important to note that, you have to decide if DFS **can pass** over the **next** cell, not the **current** cell.
 
 The most straightforward cases are noted in the problem statement (https://github.com/spiralgo/algorithms/issues/48) for the first and second examples.
 
 But, I think there is another case that is more veiled and pretty hard to see at the first glance.
 
 I tried to draw the mentioned case:
 
IMG-1626

 In the above-mentioned case; you must pass at the first encounter but you have to stop after another turn of DFS (and in a different direction.)


altayhunoglu:
image

ErdemT09:
Why is the helper method called "dps" and not "dfs"?

altayhunoglu:
> Why is the helper method called "dps" and not "dfs"?
> Does it stand for something like "depth prioritized search"?

No no. It is just a typo. Thanks for your attention. Fixing now.

src/main/java/algorithms/curated170/medium/TheMaze.java
```diff
@@ -0,0 +1,66 @@
...
+            int nextX = x + dirx;
+            int nextY = y + diry;
+
+            while (canPass(nextX, nextY)
+                    && !visited[nextX][nextY]) {
+                int prx = nextX;
+                int pry = nextY;
+                nextX += dirx;
+                nextY += diry;
+                result = result || !canPass(nextX, nextY) && dps(prx, pry);
```
ErdemT09, line 51: I think using parenthesis here for the operators would make the code more readable.
As a note for those who do not know about the precedence of the operators. 
image
Source: http://www.math.uaa.alaska.edu/~afkjm/cs110/handouts/IfStatements.pdf

altayhunoglu: Nice idea to add this table.

src/main/java/algorithms/curated170/medium/TheMaze.java
```diff
@@ -0,0 +1,66 @@
...
+            int nextX = x + dirx;
+            int nextY = y + diry;
+
+            while (canPass(nextX, nextY)
+                    && !visited[nextX][nextY]) {
+                int prx = nextX;
+                int pry = nextY;
+                nextX += dirx;
+                nextY += diry;
+                result = result || (!canPass(nextX, nextY) && dfs(prx, pry));
```
ErdemT09, line 51: Let's say that we have went left from some tile. This loop implies that we will also traverse the same from left to right until we see that the right tile is already visited. Isn't this to some degree duplicate unneeded calculation or am I seeing it wrong?

altayhunoglu: It is duplicated but I am not sure if it is unnecessary.

Somehow we have to distinguish the above-mentioned case and the case in Example-2.

It is one of the alternative ways.

We can change it when we find a more efficient way.

ErdemT09: We can perhaps put a simple statement like "Don't go right" to the method call.  We can wrap this up in an integer or a boolean value.

altayhunoglu:
I will also add BFS solution. Maybe it is much more appropriate as it can prevent dead-ends more efficiently.

altayhunoglu:
[Adds BFS, refactor DFS.](https://github.com/spiralgo/algorithms/pull/282/commits/b57f4932d61253bd64e1339836a94774c85a5b65)

altayhunoglu:
[
Adds our first unit test for The Maze - DFS solution.](https://github.com/spiralgo/algorithms/pull/282/commits/b201c6c1455b98f4d3a659f9f639788fff150311)

ErdemT09:
Here is what I meant:
```java
public boolean dfs(int x, int y, int arrivedFrom) {

        if (x == destination[0] && y == destination[1]) {
            return true;
        }

        visited[x][y] = true;
        boolean result = false;

           for (int i = 0; i<4; i++){
               
            if((i == 0 && arrivedFrom == 3)
            ||(i == 1 && arrivedFrom == 2)
            || (i == 2 && arrivedFrom == 1)
            || (i == 3 && arrivedFrom == 0))
            {
                continue;
            }
            int dirx = directions[i][0];
            int diry = directions[i][1];
            int nextX = x + dirx;
            int nextY = y + diry;

            while (canPass(nextX, nextY)
                    && !visited[nextX][nextY]) {
                int prx = nextX;
                int pry = nextY;
                nextX += dirx;
                nextY += diry;
                result = result || !canPass(nextX, nextY) && dfs(prx, pry, i);

            }
       }

        return result;
    }
```

altayhunoglu:
> Here is what I meant:
> 
 
 Nice idea Erdem, thanks.
 I am not sure what is the exact run-time efficiency as it changes in each run:
 
image

But, it is fine to add to be more specific.

ErdemT09:
In the BFS solution, we can perhaps implement the algorithm of not going back to the place we came from by changing the boolean matrix with a `int`, or simply `byte` matrix.

ErdemT09:
I think the solutions are current quite fine. We can add other settings to them in the future if needed.

ErdemT09 (approved):

PR #283, opened 2021-05-23
Adds DFS, BFS and Dijkstra's algorithm solutions for "The Maze II"

### LINKED ISSUE #56: 505. The Maze II

altayhunoglu:
resolves #56

This question is a version of https://github.com/spiralgo/algorithms/pull/282, but this time we are asked to find the shortest path to the target.

Dijkstra's algorithm appears to be the fastest with ~5ms and the way it searches is almost the same as the BFS.

**But Dijkstra starts with detecting the node which has the least distance from the starting point.**

`Priority Queue` is a perfect match to poll the closest node.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
@@ -1,0 +1,102 @@
+package algorithms.curated170.medium.maze2;
+
+import java.util.Arrays;
+import java.util.PriorityQueue;
+
+ 
+public class TheMazeIIDijkstra {
+ 
+    int[][] maze;
+    int[] destination;
+    int[][] distance;
+  
+    boolean visited[][];
+    
+    int[][] directions = {{0, -1}, {1, 0}, {-1, 0}, {0, 1}};
+    
+  public int shortestDistance(int[][] maze, int[] start, int[] dest) {
+        this.distance = new int[maze.length][maze[0].length];
+        this.destination = dest;
+        this.visited = new boolean[maze.length][maze[0].length];
+        this.maze = maze;
+
+        for (int[] row: distance)
+            Arrays.fill(row, Integer.MAX_VALUE);
+
+        distance[start[0]][start[1]] = 0;
+    
+        pq.add(new int[]{start[0], start[1], 0});
+        
+        dijkstra();
+
+        return distance[dest[0]][dest[1]] == Integer.MAX_VALUE ? -1 : distance[dest[0]][dest[1]];
+    }
+   
+    public int[] minDistance() {
+        int[] min={-1,-1};
+        int min_val = Integer.MAX_VALUE;
+        for (int i = 0; i < distance.length; i++) {
+            for (int j = 0; j < distance[0].length; j++) {
+                if (!visited[i][j] && distance[i][j] < min_val) {
+                    min = new int[] {i, j};
+                    min_val = distance[i][j];
+                }
+            }
+        }
+        return min;
+    }
+  
+    PriorityQueue<int[]> pq = new PriorityQueue<>((p1, p2) -> p1[2] - p2[2]); 
+
+    
+    public void dijkstra() {
+     
+        while (!pq.isEmpty()) {
+        
+            int[] s = pq.poll();
+           
+            if (s[0] < 0)
+                break;
+            
+            int x = s[0];
+            int y = s[1];
+                  
+            if (x == destination[0] && y == destination[1]) {
+                return;
+            }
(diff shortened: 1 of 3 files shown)


ErdemT09:
Issue: #56

altayhunoglu:
Hello @ErdemT09,
 
I have done some improvements in Dijkstra implementation.
The latest commit:
- removes `distance` array as it is obsolete after we implement PQ to get the closest node.
- removes `visited` array to improve space efficiency, utilizes the `maze` matrix for this purpose.
- removes unnecessary distance compare as PQ already does this.

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
```diff
@@ -0,0 +1,77 @@
...
+           
+            if (startPoint[0] < 0)
+                break;
+            
+            int x = startPoint[0];
+            int y = startPoint[1];
+            
+            if(maze[x][y]==-1) continue;
+            
+            maze[x][y] = -1;
```
ErdemT09, line 35: I reccommend moving -1 to a constant called "VISITED" for readability.

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
```diff
@@ -0,0 +1,77 @@
...
+        PriorityQueue<int[]> pq = new PriorityQueue<>((p1, p2) -> p1[2] - p2[2]); 
+        int[][] directions = {{0, -1}, {1, 0}, {-1, 0}, {0, 1}};
+     
+        pq.add(new int[]{start[0], start[1], 0});
+        while (!pq.isEmpty()) {
+        
+            int[] startPoint = pq.poll();
+           
+            if (startPoint[0] < 0)
+                break;
```
ErdemT09, line 28: I don't understand the function of this statement here. Can you please explain that?

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+        return -1;
+    }
+ 
+     boolean canPass(int nextX, int nextY) {
+            return (nextX >= 0 && nextY >= 0
+                    && nextX < maze.length
+                    && nextY < maze[0].length
+                    && maze[nextX][nextY] != 1);
```
ErdemT09, line 74: Why don't we check for this condition unlike in the previous two solutions?
`maze[nextX][nextY] == 0);`

ErdemT09:
I have accidentally deleted the branch. I should have been more careful.

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
ErdemT09: I have understood the reason why we don't do this.
In the previous solution, the tiles that we come across are either empty or a wall. If empty, we move through them. If wall, we stop. Here, we have a three valued logic: A tile is either empty but visited, empty or a wall. We should be able to move through an empty tile even if it is visited. At the end, we can't stop and branch off there, but whilst moving to another valid tile, we should be able to pass them.

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
ErdemT09: I still suggest doing this.

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
altayhunoglu: Thanks for the explanation Erdem. Appreciated.

src/main/java/algorithms/curated170/medium/maze2/TheMazeIIDijkstra.java
altayhunoglu: Oke, it is done. Thanks for the suggestion.

ErdemT09 (approved):
Approved. A good question in terms of graph theory.

PR #284, opened 2021-05-23
Adds Gradle Wrapper supports.

altayhunoglu:
Our project uses Gradle to build the codebase.

But we need to use Gradle Wrapper which will provide:

 - a standard way to specify and easily update the version of Gradle used across our future companies.
 - build without Gradle installed on our future audiences' local machines
 - integration with GitHub Actions and other software development automation tools.


ErdemT09 (approved):
I don't really understand how this works but this will definitely render helpful. Thank you for the valuable work.

PR #286, opened 2021-05-25
Adds Dijkstra's algorithm solutions for "The Maze III"

altayhunoglu:
resolves https://github.com/spiralgo/algorithms/issues/231

It is almost the same with: https://github.com/spiralgo/algorithms/pull/283
 
**But this time, there is a hole into which the ball can fall (instead of a destination through which the ball can pass.)**

 Nevertheless, it asked us to **print the direction labels** of the shortest path this time.

It is why I just employed the same code in "The Maze II" according to new restrictions.

I add Dijkstra's algorithm implementation using PQ. 

Even if the code is robust enough, it can be done faster utilizing other data structures.

image

### DIFF
--- a/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
@@ -1,0 +1,124 @@
+package algorithms.curated170.medium.maze3;
+
+import java.util.Comparator;
+import java.util.PriorityQueue;
+
+public class TheMazeIIIDijkstra {
+
+    int[][] maze;
+    int[] hole;
+
+    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {
+
+        this.maze = maze;
+        this.hole = hole;
+        return dijkstra(ball);
+    }
+
+    public String dijkstra(int[] ball) {
+
+        PriorityQueue<Point> pq = new PriorityQueue<>(new Comparator<Point>() {
+            @Override
+            public int compare(Point p1, Point p2) {
+                if (p1.distance == p2.distance) {
+                    return p1.path.compareTo(p2.path);
+                } else {
+                    return p1.distance - p2.distance;
+                }
+            }
+        });
+        int[][] directions = {{0, -1}, {1, 0}, {-1, 0}, {0, 1}};
+        char[] directionSymbols = {'l', 'd', 'u', 'r'};
+
+        pq.add(new Point(ball[0], ball[1], 0, ""));
+
+        while (!pq.isEmpty()) {
+
+            Point startPoint = pq.poll();
+
+            if (startPoint.x < 0) {
+                break;
+            }
+
+            int x = startPoint.x;
+            int y = startPoint.y;
+
+            if (maze[x][y] == -1) {
+                continue;
+            }
+
+            maze[x][y] = -1;
+
+            if (isInTheHall(x, y)) {
+                return startPoint.path;
+            }
+
+            for (int i = 0; i < 4; i++) {
+                int dirx = directions[i][0];
+                int diry = directions[i][1];
+
+                int nextX = x;
+                int nextY = y;
+
+                int currentDistance = startPoint.distance;
+
+                while (canPass(nextX, nextY) && !isInTheHall(nextX, nextY)) {
+                    nextX += dirx;
+                    nextY += diry;
+                    currentDistance++;
+                }
+
+                if (!isInTheHall(nextX, nextY)) {
+                    nextX -= dirx;
+                    nextY -= diry;
+                    currentDistance--;
+                }
+
+                if (maze[nextX][nextY] != -1) {
+                    String path = startPoint.path + directionSymbols[i];
+                    pq.add(new Point(nextX, nextY, currentDistance, path));
+                }
+            }
+        }
+        return "impossible";
+    }
+
+    boolean isInTheHall(int x, int y) {
+        return (x == hole[0] && y == hole[1]);
+    }
+
+    boolean canPass(int nextX, int nextY) {
+        return (nextX >= 0 && nextY >= 0
+                && nextX < maze.length
+                && nextY < maze[0].length
+                && maze[nextX][nextY] != 1);
+
+    }
+
+    private class Point {
+
+        private int x, y, distance;
+        private String path;
+
+        Point(int x, int y, int distance, String path) {
+            this.x = x;
+            this.y = y;
+            this.distance = distance;
+            this.path = path;
+
+        }
+    }
+
+    public static void main(String[] args) {
+        TheMazeIIIDijkstra theMaze = new TheMazeIIIDijkstra();
+        int[][] maze = {{0, 0, 1, 0, 0},
+        {0, 0, 0, 0, 0},
+        {0, 0, 0, 1, 0},
+        {1, 1, 0, 1, 1},
+        {0, 0, 0, 0, 0}};
+
+        int[] start = {0, 4}, destination = {4, 4};
+        String result = theMaze.findShortestWay(maze, start, destination);
+        System.out.println(result);
+    }
+}
--- a/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
@@ -35,11 +35,7 @@ public class TheMazeIIIDijkstra {
         while (!pq.isEmpty()) {
 
             Point startPoint = pq.poll();
-
-            if (startPoint.x < 0) {
-                break;
-            }
-
+            
             int x = startPoint.x;
             int y = startPoint.y;
 
@@ -49,7 +45,7 @@ public class TheMazeIIIDijkstra {
 
             maze[x][y] = -1;
 
-            if (isInTheHall(x, y)) {
+            if (isInTHole(x, y)) {
                 return startPoint.path;
             }
 
@@ -62,7 +58,7 @@ public class TheMazeIIIDijkstra {
 
                 int currentDistance = startPoint.distance;
 
-                while (canPass(nextX, nextY) && !isInTheHall(nextX, nextY)) {
+                while (canPass(nextX, nextY) && !isInTHole(nextX, nextY)) {
                     nextX += dirx;
                     nextY += diry;
                     currentDistance++;
@@ -83,7 +79,7 @@ public class TheMazeIIIDijkstra {
         return "impossible";
     }
 
-    boolean isInTheHall(int x, int y) {
+    boolean isInTHole(int x, int y) {
         return (x == hole[0] && y == hole[1]);
     }
 
--- a/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
@@ -4,7 +4,6 @@ import java.util.Comparator;
 import java.util.PriorityQueue;
 
 public class TheMazeIIIDijkstra {
-
     int[][] maze;
     int[] hole;
 
--- a/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
@@ -64,7 +64,7 @@ public class TheMazeIIIDijkstra {
                     currentDistance++;
                 }
 
-                if (!isInTheHall(nextX, nextY)) {
+                if (!isInTHole(nextX, nextY)) {
                     nextX -= dirx;
                     nextY -= diry;
                     currentDistance--;
--- a/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
@@ -41,7 +41,7 @@ public class TheMazeIIIDijkstra {
 
             maze[x][y] = -1;
 
-            if (isInTHole(x, y)) {
+            if (isInTheHole(x, y)) {
                 return startPoint.path.toString();
             }
 
@@ -54,13 +54,13 @@ public class TheMazeIIIDijkstra {
 
                 int currentDistance = startPoint.distance;
 
-                while (canPass(nextX, nextY) && !isInTHole(nextX, nextY)) {
+                while (canPass(nextX, nextY) && !isInTheHole(nextX, nextY)) {
                     nextX += dirx;
                     nextY += diry;
                     currentDistance++;
                 }
 
-                if (!isInTHole(nextX, nextY)) {
+                if (!isInTheHole(nextX, nextY)) {
                     nextX -= dirx;
                     nextY -= diry;
                     currentDistance--;
@@ -78,7 +78,7 @@ public class TheMazeIIIDijkstra {
         return "impossible";
     }
 
-    boolean isInTHole(int x, int y) {
+    boolean isInTheHole(int x, int y) {
         return (x == hole[0] && y == hole[1]);
     }
 
--- a/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
+++ b/src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
@@ -1,9 +1,4 @@
-package algorithms.curated170.medium.maze3;
-
-import java.util.Comparator;
-import java.util.PriorityQueue;
-
-public class TheMazeIIIDijkstra {
+public class Solution {
     int[][] maze;
     int[] hole;
 
@@ -26,15 +21,15 @@ public class TheMazeIIIDijkstra {
                 }
             }
         });
-        int[][] directions = {{0, -1}, {1, 0}, {-1, 0}, {0, 1}};
-        char[] directionSymbols = {'l', 'd', 'u', 'r'};
+        int[][] directions = { { 0, -1 }, { 1, 0 }, { -1, 0 }, { 0, 1 } };
+        char[] directionSymbols = { 'l', 'd', 'u', 'r' };
 
-        pq.add(new Point(ball[0], ball[1], 0, ""));
+        pq.add(new Point(ball[0], ball[1], 0, new StringBuilder("")));
 
         while (!pq.isEmpty()) {
 
             Point startPoint = pq.poll();
-            
+
             int x = startPoint.x;
             int y = startPoint.y;
 
@@ -45,7 +40,7 @@ public class TheMazeIIIDijkstra {
             maze[x][y] = -1;
 
             if (isInTHole(x, y)) {
-                return startPoint.path;
+                return startPoint.path.toString();
             }
 
             for (int i = 0; i < 4; i++) {
@@ -70,7 +65,10 @@ public class TheMazeIIIDijkstra {
                 }
 
                 if (maze[nextX][nextY] != -1) {
-                    String path = startPoint.path + directionSymbols[i];
+
+                    StringBuilder path = new StringBuilder(startPoint.path);
+                    path.append(directionSymbols[i]);
+
                     pq.add(new Point(nextX, nextY, currentDistance, path));
                 }
             }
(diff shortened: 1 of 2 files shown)


ErdemT09:
> Even if the code is robust enough, it can be done faster utilizing other data structures.
What could be those other data structures? Special types of heaps like the Fibonacii heap or such?

src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
```diff
@@ -0,0 +1,124 @@
...
+                if (maze[nextX][nextY] != -1) {
+                    String path = startPoint.path + directionSymbols[i];
+                    pq.add(new Point(nextX, nextY, currentDistance, path));
+                }
+            }
+        }
+        return "impossible";
+    }
+
+    boolean isInTheHall(int x, int y) {
```
ErdemT09, line 86: I guess a typo.
```suggestion
    boolean isInTheHole(int x, int y) {
```

src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
```diff
@@ -0,0 +1,124 @@
...
+
+        pq.add(new Point(ball[0], ball[1], 0, ""));
+
+        while (!pq.isEmpty()) {
+
+            Point startPoint = pq.poll();
+
+            if (startPoint.x < 0) {
+                break;
+            }
```
ErdemT09, line 41: As in the previous question, I don't understand the function of this statement.

ErdemT09: > As in the previous question, I don't understand the function of this statement. Why do we check x but bot y?

So, does it work without it?

altayhunoglu: Yeah, it was just a dirty residue of a code trial. It is my fault. Sorry about that.

altayhunoglu:
> > Even if the code is robust enough, it can be done faster utilizing other data structures.
> 
> What could be those other data structures? Special types of heaps like the Fibonacci heap or such?

An ordinary array may affect the efficiency and takes ~2ms.
As it is not readable enough, we can add it just as a note here; with an intention for future refactoring.

```java
class Solution {
    int min; 
    String minS; 
    int[] hole;
    int[][] maze; 
    int[][] map; 
    int[][] dirs = {{0, 1}, {-1, 0}, {1, 0}, {0, -1}}; 
    
    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {
        this.min = Integer.MAX_VALUE; 
        this.minS = null;
        this.hole = hole; 
        this.maze = maze; 
        this.map = new int[maze.length][maze[0].length];
        for (int i = 0; i < map.length; i++) Arrays.fill(map[i], Integer.MAX_VALUE); 
        
        move(ball[0], ball[1], 0, "", -1);
        return (minS == null) ? "impossible" : minS;
    }
    
    private void move(int r, int c, int cnt, String path, int dir){
        if (cnt > min || cnt > map[r][c]) return;
        if (dir != -1){
            if(dir == 0) path += 'r';
            else if(dir == 1) path += 'u';
            else if(dir == 2) path += 'd';
            else path += 'l';
    
            while (r >= 0 && r < maze.length && c >= 0 && c < maze[0].length && maze[r][c] == 0){
                map[r][c] = Math.min(map[r][c], cnt); 
                if (r == hole[0] && c == hole[1]){
                    if (cnt == min && path.compareTo(minS) < 0){
                        minS = path;
                    } else if(cnt < min){
                        min = cnt; 
                        minS = path; 
                    }
                    return; 
                }
                r += dirs[dir][0];
                c += dirs[dir][1];
                cnt++;
            }
            r -= dirs[dir][0];
            c -= dirs[dir][1];
            cnt--;
        }
        
        for (int i = 0; i < dirs.length; i++){
            if(dir == i) continue;
            if(dir == (3 - i)) continue;
            int newR = r + dirs[i][0];
            int newC = c + dirs[i][1];
            if (newR >= 0 && newR < maze.length && newC >= 0 && newC < maze[0].length && maze[newR][newC] == 0){
                move(r, c, cnt, path, i);
            }
        }
    }
}
```

ErdemT09:
That code does around the same thing we had done for the DFS solution in Maze II. I personally haven't noticed anything more remarkable than that at the first glance.

ErdemT09:
The PR can remain open. I will try implementing this DFS algorithm in a more simple way. It shouldn't be that hard as we have done it twice already.

ErdemT09:
This code can actually be sped up using few modifications and perhaps some String Builders. However, here is the original code without any semantical changes, with comments for current explanation and refactoring. 

```java
public class TheMazeIIIDFS {
    int min;
    String minS;
    int[] hole;
    int[][] maze;
    int[][] map;
    final int[][] directions = { { 0, 1 }, { -1, 0 }, { 1, 0 }, { 0, -1 } };
    final char[] directionSymbols = { 'r', 'u', 'd', 'l' };

    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {
        this.min = Integer.MAX_VALUE;
        this.minS = null;
        this.hole = hole;
        this.maze = maze;
        this.map = new int[maze.length][maze[0].length];
        for (int i = 0; i < map.length; i++)
            Arrays.fill(map[i], Integer.MAX_VALUE);

        // At the beginning, we are at the given position, where path length = 0, path
        // string is empty, and our initial direction is none(-1)
        move(ball[0], ball[1], 0, "", -1);
        return (minS == null) ? "impossible" : minS;
    }

    private void move(int x, int y, int cnt, String path, int dir) {
        if (cnt > min || cnt > map[x][y])
            return;

        if (dir != -1) {

            // values of dir correspond to various chars, we add these chars in relation to
            // the direction we came from
            path += directionSymbols[dir];

            while (canPass(x, y)) {
                map[x][y] = Math.min(map[x][y], cnt);

                if (checkTargetReached(x, y, cnt, path)) {
                    return;
                }
                x += directions[dir][0];
                y += directions[dir][1];
                cnt++;
            }
            x -= directions[dir][0];
            y -= directions[dir][1];
            cnt--;
        }

        moveToFourDirections(x, y, cnt, path, dir);
    }

    private boolean canPass(int x, int y) {
        return x >= 0 && x < maze.length && y >= 0 && y < maze[0].length && maze[x][y] == 0;
    }

    private boolean checkTargetReached(int x, int y, int cnt, String path) {
        if (x == hole[0] && y == hole[1]) {

            // if path lengths are equal and the new one is lexographically smaller
            if (cnt == min && path.compareTo(minS) < 0) {
                minS = path;
            }
            // if the path length is smaller
            else if (cnt < min) {
                min = cnt;
                minS = path;
            }
            return true;
        }

        return false;
    }

    private void moveToFourDirections(int x, int y, int cnt, String path, int dir) {
        for (int i = 0; i < 4; i++) {

            /*
             * This doesn't effect the result of the algorithm bu it makes it faster,
             * because: It doesn't make sense to go in the direction by which we arrived
             * from. For example, if we have arrived from right, we can't go further right,
             * because we have hit a wall and that is why we have stopped at this tile in
             * the first place.
             */
            if (dir == i)
                continue;

            /*
             * Second, this line tries to do something similar but in an other way. The only
             * possible way this if statement can be true is that dir == 0 and i == 3 In
             * other words, if we came from right and are trying to go left
             */

            if (dir == (3 - i))
                continue;

            /*
             * Unlike in our previous solutions of Maze 2 and Maze 1, we move to the valid
             * adjacenet tiles of a stable vertex, not to another stable vertex
             */
            int newR = x + directions[i][0];
            int newC = y + directions[i][1];
            if (canPass(newR, newC)) {
                move(x, y, cnt, path, i);
            }
        }
    }
```

ErdemT09:
String Builder can perhaps also be used in the Dijkstra solution.

ErdemT09:
> String Builder can perhaps also be used in the Dijkstra solution.
Which looks like this:

```java
public class TheMazeIIIDijkstra {
    int[][] maze;
    int[] hole;

    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {

        this.maze = maze;
        this.hole = hole;
        return dijkstra(ball);
    }

    public String dijkstra(int[] ball) {

        PriorityQueue<Point> pq = new PriorityQueue<>(new Comparator<Point>() {
            @Override
            public int compare(Point p1, Point p2) {
                if (p1.distance == p2.distance) {
                    return p1.path_.compareTo(p2.path_);
                } else {
                    return p1.distance - p2.distance;
                }
            }
        });
        int[][] directions = { { 0, -1 }, { 1, 0 }, { -1, 0 }, { 0, 1 } };
        char[] directionSymbols = { 'l', 'd', 'u', 'r' };

        pq.add(new Point(ball[0], ball[1], 0, new StringBuilder("")));

        while (!pq.isEmpty()) {

            Point startPoint = pq.poll();

            int x = startPoint.x;
            int y = startPoint.y;

            if (maze[x][y] == -1) {
                continue;
            }

            maze[x][y] = -1;

            if (isInTHole(x, y)) {
                return startPoint.path_.toString();
            }

            for (int i = 0; i < 4; i++) {
                int dirx = directions[i][0];
                int diry = directions[i][1];

                int nextX = x;
                int nextY = y;

                int currentDistance = startPoint.distance;

                while (canPass(nextX, nextY) && !isInTHole(nextX, nextY)) {
                    nextX += dirx;
                    nextY += diry;
                    currentDistance++;
                }

                if (!isInTHole(nextX, nextY)) {
                    nextX -= dirx;
                    nextY -= diry;
                    currentDistance--;
                }

                if (maze[nextX][nextY] != -1) {

                    StringBuilder path_ = new StringBuilder(startPoint.path_);
                    path_.append(directionSymbols[i]);

                    pq.add(new Point(nextX, nextY, currentDistance, path_));
                }
            }
        }
        return "impossible";
    }

    boolean isInTHole(int x, int y) {
        return (x == hole[0] && y == hole[1]);
    }

    boolean canPass(int nextX, int nextY) {
        return (nextX >= 0 && nextY >= 0 && nextX < maze.length && nextY < maze[0].length && maze[nextX][nextY] != 1);

    }

    private class Point {

        private int x, y, distance;
        private StringBuilder path_;

        Point(int x, int y, int distance, StringBuilder path_) {
            this.x = x;
            this.y = y;
            this.distance = distance;
            this.path_ = path_;
        }
    }
```

altayhunoglu:
> > String Builder can perhaps also be used in the Dijkstra solution.
 
 It is a nice idea. It speeds up from 12ms to 6ms.
 I will change `path_` to `path ` and add this version.

altayhunoglu:
> This code can actually be sped up using few modifications and perhaps some String Builders. However, here is the original code without any semantical changes, with comments for current explanation and refactoring.


Thanks for the labor Erdem. It takes 11 ms.
Nice implementation.
image

ErdemT09:
> > This code can actually be sped up using few modifications and perhaps some String Builders. However, here is the original code without any semantical changes, with comments for current explanation and refactoring.
> 
> 
> Thanks for the labor Erdem. It takes 11 ms.
> Nice implementation.
> image
> 
> 

The original one took 2 ms. I didn't change anything as far as I'm aware of it. What could have slowed it down?

ErdemT09:
Here is the newer version using the the StringBuilder class of Java.

```java
public class TheMazeIIIDFS {
    int min;
    StringBuilder sb;
    int[] hole;
    int[][] maze;
    int[][] map;
    final int[][] directions = { { 0, 1 }, { -1, 0 }, { 1, 0 }, { 0, -1 } };
    final char[] directionSymbols = { 'r', 'u', 'd', 'l' };

    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {
        this.min = Integer.MAX_VALUE;
        this.sb = null;
        this.hole = hole;
        this.maze = maze;
        this.map = new int[maze.length][maze[0].length];
        for (int i = 0; i < map.length; i++)
            Arrays.fill(map[i], Integer.MAX_VALUE);

        // At the beginning, we are at the given position, where path length = 0, path
        // string is empty, and our initial direction is none(-1)
        move(ball[0], ball[1], 0, new StringBuilder(""), -1);
        return (sb == null) ? "impossible" : sb.toString();
    }

    private void move(int x, int y, int cnt, StringBuilder path_, int dir) {
        if (cnt > min || cnt > map[x][y])
            return;

        StringBuilder newPath = new StringBuilder(path_);
        if (dir != -1) {

            // values of dir correspond to various chars, we add these chars in relation to
            // the direction we came from
            newPath.append(directionSymbols[dir]);

            while (canPass(x, y)) {
                map[x][y] = Math.min(map[x][y], cnt);

                if (checkTargetReached(x, y, cnt, newPath)) {
                    return;
                }
                x += directions[dir][0];
                y += directions[dir][1];
                cnt++;
            }
            x -= directions[dir][0];
            y -= directions[dir][1];
            cnt--;
        }

        moveToFourDirections(x, y, cnt, newPath, dir);
    }

    private boolean canPass(int x, int y) {
        return x >= 0 && x < maze.length && y >= 0 && y < maze[0].length && maze[x][y] == 0;
    }

    private boolean checkTargetReached(int x, int y, int cnt, StringBuilder path_) {
        if (x == hole[0] && y == hole[1]) {

            // if path lengths are equal and the new one is lexographically smaller
            if (cnt == min && path_.compareTo(sb) < 0) {
                sb = path_;
            }
            // if the path length is smaller
            else if (cnt < min) {
                min = cnt;
                sb = path_;
            }
            return true;
        }

        return false;
    }

    private void moveToFourDirections(int x, int y, int cnt, StringBuilder path_, int dir) {
        for (int i = 0; i < 4; i++) {

            /*
             * This doesn't effect the result of the algorithm bu it makes it faster,
             * because: It doesn't make sense to go in the direction by which we arrived
             * from. For example, if we have arrived from right, we can't go further right,
             * because we have hit a wall and that is why we have stopped at this tile in
             * the first place.
             */
            if (dir == i)
                continue;

            /*
             * Second, this line tries to do something similar but in an other way. The only
             * possible way this if statement can be true is that dir == 0 and i == 3 In
             * other words, if we came from right and are trying to go left
             */

            if (dir == (3 - i))
                continue;

            /*
             * Unlike in our previous solutions of Maze 2 and Maze 1, we move to the valid
             * adjacenet tiles of a stable vertex, not to another stable vertex
             */
            int newR = x + directions[i][0];
            int newC = y + directions[i][1];
            if (canPass(newR, newC)) {
                move(x, y, cnt, path_, i);
            }
        }
    }
```

ErdemT09:
The direction check we had done in #283 can also be implemented here as such. We could perhaps try running like this as well. However, I doubt that it would change the efficiency because the sanity checks in the move method probably already handle this.

```java
    private void moveToFourDirections(int x, int y, int cnt, StringBuilder path, int dir) {
        for (int i = 0; i < 4; i++) {

            if (isInvalidDirection(dir, i))
                continue;
                
            /*
             * Unlike in our previous solutions of Maze 2 and Maze 1, we move to the valid
             * adjacenet tiles of a stable vertex, not to another stable vertex
             */
            int newR = x + directions[i][0];
            int newC = y + directions[i][1];
            if (canPass(newR, newC)) {
                move(x, y, cnt, path, i);
            }
        }
    }

    private boolean isInvalidDirection(int dir, int dest)
    {
        return (dir == dest 
                || (dir == 0 && dest == 3)
                || (dir == 1 && dest == 2)
                || (dir == 3 && dest == 0)
                || (dir == 2 && dest == 1));
    }
```

src/main/java/algorithms/curated170/medium/maze3/TheMazeIIIDijkstra.java
```diff
@@ -0,0 +1,103 @@
...
+                    path.append(directionSymbols[i]);
+
+                    pq.add(new Point(nextX, nextY, currentDistance, path));
+                }
+            }
+        }
+        return "impossible";
+    }
+
+    boolean isInTHole(int x, int y) {
```
ErdemT09, line 81: I guess a typo.

altayhunoglu: Thank you for your attention Erdem. Fixed.

ErdemT09:
> The direction check we had done in #283 can also be implemented here..

If you haven't seen it, I am noting it once again.

altayhunoglu:
> > The direction check we had done in #283 can also be implemented here..
> 
> If you haven't seen it, I am noting it once again.


Thank you, Erdem.
 I will have a look at it after I grasp the intuition for your Factor combination solution.

altayhunoglu:
> > > The direction check we had done in #283 can also be implemented here..
> > 
> > 
> > If you haven't seen it, I am noting it once again.
> 
> Thank you, Erdem.
> I will have a look at it after I grasp truly the intuition for your Factor combination solution.

I am not sure why but I increased the run-time.

image

ErdemT09:
Perhaps simplifiying all to this would be the best.
```java
    private void moveToFourDirections(int x, int y, int cnt, StringBuilder path, int dir) {
        for (int i = 0; i < 4; i++) {

            if (dir == i)
                continue;

            int newR = x + directions[i][0];
            int newC = y + directions[i][1];
            if (canPass(newR, newC)) {
                move(x, y, cnt, path, i);
            }
        }
    }
```

altayhunoglu:
> Perhaps simplifiying all to this would be the best.
> 
> ```java
>     private void moveToFourDirections(int x, int y, int cnt, StringBuilder path, int dir) {
>         for (int i = 0; i < 4; i++) {
> 
>             if (dir == i)
>                 continue;
> 
>             int newR = x + directions[i][0];
>             int newC = y + directions[i][1];
>             if (canPass(newR, newC)) {
>                 move(x, y, cnt, path, i);
>             }
>         }
>     }
> ```

image


This is because we execute the if-condition for each iteration.

ErdemT09:
Lastly, we can try going without such checks at all.
```java
    private void moveToFourDirections(int x, int y, int cnt, StringBuilder path, int dir) {
        for (int i = 0; i < 4; i++) {

            if (dir == i)
                continue;

            int newR = x + directions[i][0];
            int newC = y + directions[i][1];
            if (canPass(newR, newC)) {
                move(x, y, cnt, path, i);
            }
        }
    }
```

altayhunoglu:
> Lastly, we can try going without such checks at all.
 
 Yep. The checks are good.
 
image

ErdemT09:
The problem is then:
Why is the two small checks that we have in the solution the most optimum of  all types of checks?

altayhunoglu:
`(dir == (3 - i))` almost equals to

```java
((dir == 0 && dest == 3)
                || (dir == 1 && dest == 2)
                || (dir == 3 && dest == 0)
                || (dir == 2 && dest == 1));
```

ErdemT09:
> `(dir == (3 - i))` almost equals to

I have noticed that more thoroughly now. That actually makes sense.

ErdemT09:
I will extract that small part to its method as well and remove all comments in the code then. Then, we are done with this question, I think. Handling this question was a good collaboration though.

altayhunoglu:
> I will extract that small part to its method as well and remove all comments in the code then. Then, we are done with this question, I think. Handling this question was a good collaboration though.

As it is always a pleasure to work with you.
Erdem, thanks.

ErdemT09:
```java
    int min;
    StringBuilder minS;
    int[] hole;
    int[][] maze;
    int[][] map;
    final int[][] directions = { { 0, 1 }, { -1, 0 }, { 1, 0 }, { 0, -1 } };
    final char[] directionSymbols = { 'r', 'u', 'd', 'l' };

    public String findShortestWay(int[][] maze, int[] ball, int[] hole) {
        this.min = Integer.MAX_VALUE;
        this.minS = null;
        this.hole = hole;
        this.maze = maze;
        this.map = new int[maze.length][maze[0].length];

        for (int[] row : map) {
            Arrays.fill(row, Integer.MAX_VALUE);
        }

        move(ball[0], ball[1], 0, new StringBuilder(""), -1);

        return (minS == null) ? "impossible" : minS.toString();
    }

    private void move(int x, int y, int cnt, StringBuilder path, int dir) {
        if (cnt > min || cnt > map[x][y]) {
            return;
        }

        StringBuilder newPath = new StringBuilder(path);
        if (dir != -1) {

            newPath.append(directionSymbols[dir]);

            while (canPass(x, y)) {
                map[x][y] = Math.min(map[x][y], cnt);

                if (checkTargetReached(x, y, cnt, newPath)) {
                    return;
                }
                x += directions[dir][0];
                y += directions[dir][1];
                cnt++;
            }
            x -= directions[dir][0];
            y -= directions[dir][1];
            cnt--;
        }

        moveToFourDirections(x, y, cnt, newPath, dir);
    }

    private boolean canPass(int x, int y) {
        return x >= 0 && x < maze.length && y >= 0 && y < maze[0].length && maze[x][y] == 0;
    }

    private boolean checkTargetReached(int x, int y, int cnt, StringBuilder path) {
        if (x == hole[0] && y == hole[1]) {

            if (cnt == min && path.compareTo(minS) < 0) {
                minS = path;
            } else if (cnt < min) {
                min = cnt;
                minS = path;
            }
            return true;
        }

        return false;
    }

    private void moveToFourDirections(int x, int y, int cnt, StringBuilder path, int dir) {
        for (int i = 0; i < 4; i++) {

            if (isInvalidDestination(dir, i)) {
                continue;
            }

            int newR = x + directions[i][0];
            int newC = y + directions[i][1];
            if (canPass(newR, newC)) {
                move(x, y, cnt, path, i);
            }
        }
    }

    private boolean isInvalidDestination(int dir, int dest) {
        return (dir == dest || (dir == (3 - dest)));
    }
```

altayhunoglu:
Oke, we removed the comments.

ErdemT09 (approved):
All done here then, solving such problems together is always valuable.
Approved.

PR #287, opened 2021-05-26
Adds DFS, Union-Find Array, Union-Find HashMap solutions (Sentence Similarity II)

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/66

When we represent the connections of the word pairs as a `graph`, it becomes easier to solve the problem using either DFS or Union-Find.

In the algorithm, we will mainly ask two questions to decide if the given sentences are similar:
1- Do the sentences have the same number of words?
2- Are the words (at the same index of the given sentences) similar?
    - If we use `DFS`, this means "Does the pair graph have a cycle for these two words?"
    - If we use `Union-Find`, this means "Are these two words in the same group with the same parent?"
    
    

> For example, sentence1= ["great", "acting", "skills"] and sentence2= ["fine", "drama", "talent"] are similar, if the similar word pairs are similar = [["great", "good"], ["fine", "good"], ["acting","drama"], ["skills","talent"]].

    
 
IMG-1635

### DIFF
--- a/src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIDFS.java
+++ b/src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIDFS.java
@@ -1,0 +1,58 @@
+package algorithms.curated170.medium.sentencesimilarity2;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
+public class SentenceSimilarityIIDFS {
+
+    Map<String, List<String>> graph;
+    HashSet<String> visited;
+
+    public boolean areSentencesSimilarTwo(String[] words1, String[] words2, List<List<String>> pairs) {
+        if (words1.length != words2.length) {
+            return false;
+        }
+        graph = new HashMap();
+
+        for (List<String> pair : pairs) {
+            for (String p : pair) {
+                if (!graph.containsKey(p)) {
+                    graph.put(p, new ArrayList());
+                }
+            }
+            graph.get(pair.get(0)).add(pair.get(1));
+            graph.get(pair.get(1)).add(pair.get(0));
+        }
+        for (int i = 0; i < words1.length; ++i) {
+            visited = new HashSet();
+            String w1 = words1[i], w2 = words2[i];
+            if (!dfs(w1, w2)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean dfs(String w1, String w2) {
+
+        visited.add(w2);
+        if (w1.equals(w2)) {
+            return true;
+        }
+
+        if (graph.containsKey(w2)) {
+            List<String> connectedWords = graph.get(w2);
+            for (String word : connectedWords) {
+                if (!visited.contains(word) && dfs(w1, word)) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+}
--- a/src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIDFS.java
+++ b/src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIDFS.java
@@ -19,9 +19,7 @@ public class SentenceSimilarityIIDFS {
 
         for (List<String> pair : pairs) {
             for (String p : pair) {
-                if (!graph.containsKey(p)) {
-                    graph.put(p, new ArrayList());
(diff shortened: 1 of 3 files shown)


ErdemT09:
In the image, drawing the graph arrows undirected or double sided would have been more accurate, because if a is similar to b, then b is similar to a because of symmetry relation.

src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIDFS.java
```diff
@@ -0,0 +1,58 @@
...
+        if (words1.length != words2.length) {
+            return false;
+        }
+        graph = new HashMap();
+
+        for (List<String> pair : pairs) {
+            for (String p : pair) {
+                if (!graph.containsKey(p)) {
+                    graph.put(p, new ArrayList());
+                }
```
ErdemT09, line 24: Small refactoring.
```suggestion
               graph.putIfAbsent(p, new ArrayList());
```

src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIUnionFindArray.java
```diff
@@ -0,0 +1,52 @@
...
+ 
+        Map<String, Integer> index = new HashMap();
+        int count = 0;
+        DSU dsu = new DSU(2 * pairs.size());
+        
+        for (List<String> pair : pairs) {
+            for (String p: pair){
+              if (!index.containsKey(p)) {
+                index.put(p, count++);
+            }
```
ErdemT09, line 22: Again, putIfAbsent() method looks simpler.

src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIUnionFindMap.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        return find(parentWord);
+    }
+
+    void union(String w1, String w2) {
+        String parentWord1 = find(w1);
+        String parentWord2 = find(w2);
+ 
+        if (!parentWord1.equals(parentWord2)) {
+            parent.put(find(w1), find(w2));
```
ErdemT09, line 52: Isn't this better in order not to repeat the process?
parent.put(parentWord2, parentWord2);

src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIUnionFindMap.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    String find(String w1) {
+        if(!parent.containsKey(w1)) return w1;
+        String parentWord = parent.get(w1);
+        if (parentWord.equals(w1)) {
+            return w1;
+        }
+        return find(parentWord);
+    }
```
ErdemT09, line 45: We can perhaps apply parent compression here to reduce the repetiton of operations:
```suggestion
    String find(String w1) {
        if(!parent.containsKey(w1)) return w1;
        String parentWord = parent.get(w1);
        if (parentWord.equals(w1)) {
            return w1;
        }
        parent.put(w1, find(parenWord);
        return parent.get(w1);
    }
```

ErdemT09:
```suggestion
    String find(String w1) {
        if(!parent.containsKey(w1)) return w1;
        String parentWord = parent.get(w1);
        if (!parentWord.equals(w1)) {
            parent.put(w1, find(parentWord));
        }
        return parent.get(w1);
    }
```

altayhunoglu:
> In the image, drawing the graph arrows undirected or double sided would have been more accurate, because if a is similar to b, then b is similar to a because of symmetry relation.

The image has been updated with bidirectional arrows.
Thanks for the attention Erdem.

src/main/java/algorithms/curated170/medium/sentencesimilarity2/SentenceSimilarityIIUnionFindMap.java
altayhunoglu: Oke these are done.

ErdemT09:
Basically a graph question perhaps covered as a string question. Thinking that way, questions like Longest Common Substring etc. are also graph questions to some degree. Anyway, approved.

ErdemT09 (approved):

PR #288, opened 2021-05-27
Adds Array and PQ solutions (253. Meeting Rooms II)

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/64

I think the simplest intuition to solve this question might be like that:

 - If there is just 1 meeting, we would need just 1 room.
 - Even if there is more than 1 meeting; if the next meetings start just after the end of the previous ones, then again we would need just 1 room. 
 - If this is not the case, we can try to use the same room for one or more meetings.

  `Conclusion`: We can use a vacated room `for any meeting` that is supposed to be held after `the meeting which ended first` (namely, the interval which has the minimum end.)
  
  It is why we need to keep track of already ended meetings and somehow poll the minimum one.
  
  PQ is a perfect match for that purpose. However, if the same idea is implemented using an array, it will be more efficient.
  
  Therefore, I add both implementations.
  
For the example below, we would need one room for [0,30] and an extra room for the others. 
We would need 2 rooms at the end.

IMG-1630

### DIFF
--- a/src/main/java/algorithms/curated170/medium/meetingrooms2/MeetingRoomsIIArray.java
+++ b/src/main/java/algorithms/curated170/medium/meetingrooms2/MeetingRoomsIIArray.java
@@ -1,0 +1,33 @@
+package algorithms.curated170.medium.meetingrooms2;
+
+import java.util.Arrays; 
+
+ 
+public class MeetingRoomsIIArray {
+     public int minMeetingRooms(int[][] intervals) {
+        int[] startTimes = new int[intervals.length];
+        int[] endTimes = new int[intervals.length];
+        
+        for (int i = 0; i < intervals.length; i++) {
+            startTimes[i] = intervals[i][0];
+            endTimes[i] = intervals[i][1];
+        }
+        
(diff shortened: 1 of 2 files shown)


ErdemT09:
I have also tried illustrating the Array solution with an example.
Here is the video: https://youtu.be/1ZTa18zkKCs

ErdemT09 (approved):

altayhunoglu:
> I have also tried illustrating the Array solution with an example.
> Here is the video: https://youtu.be/1ZTa18zkKCs

Thank you for your valuable collaboration Erdem. It is very helpful to grasp.

ErdemT09:
> I have also tried illustrating the Array solution with an example.
> Here is the video: https://youtu.be/1ZTa18zkKCs

I should have made the video a bit slower though. Here is the new version:

https://user-images.githubusercontent.com/63192680/120004049-fd679100-bfde-11eb-924a-0d28b5dd0bc7.mp4

altayhunoglu:
Yep. But not a big trouble for YouTube. I watched it with 0.25x.

PR #289, opened 2021-05-28
250. Count Univalue Subtrees

ErdemT09:
What are the conditions that a node comprises a univalue subtree?
Either one of the two must be valid:
1. It does not have child node.
2. It does have child nodes, its value is the same as those of the child nodes, its children are all univalue subtrees.

Obviously, this implies a recursive, DFS algorithm. We can do this in 2 ways:
1. If the node has children, we first compare its value with each node and then send the recursive method. If it does not have any children, a boolean value like `isUnival` should be declared `true`, so it remains so when no node is checked.
2. We pass the parent values to the children. If the child is null or, if it is a univalue subtree itself and the passed value is equal to the child's value, the methods returns true.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/countunivaluesubtrees/CountUnivalueSubtreesDFS.java
+++ b/src/main/java/algorithms/curated170/medium/countunivaluesubtrees/CountUnivalueSubtreesDFS.java
@@ -18,10 +18,10 @@ public class CountUnivalueSubtreesDFS {
 
         boolean isUnival = true;
         if (root.left != null) {
-            isUnival = root.left.val == root.val && isUnivalSubtree(root.left);
+            isUnival = isUnivalSubtree(root.left) && root.left.val == root.val;
         }
-        if (root.left != null && isUnival) {
-            isUnival = root.right.val == root.val && isUnivalSubtree(root.right);
+        if (root.left != null) {
+            isUnival = isUnival & isUnivalSubtree(root.right) && root.right.val == root.val;
         }
         if (isUnival) {
             count++;
--- a/src/main/java/algorithms/curated170/medium/countunivaluesubtrees/CountUnivalueSubtreesDFS.java
+++ b/src/main/java/algorithms/curated170/medium/countunivaluesubtrees/CountUnivalueSubtreesDFS.java
@@ -1,15 +1,17 @@
-package algorithms.curated170.medium;
+package algorithms.curated170.countunivaluesubtrees.medium;
 
 import algorithms.datastructures.TreeNode;
 
-public class CountUnivalueSubtrees {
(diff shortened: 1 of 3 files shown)


altayhunoglu:
Issue has been updated with images and the default code:
https://github.com/spiralgo/algorithms/issues/45

src/main/java/algorithms/curated170/medium/countunivaluesubtrees/CountUnivalueSubtreesDFS.java
```diff
@@ -0,0 +1,33 @@
...
+        return count;
+    }
+
+    private boolean isUnivalSubtree(TreeNode root) {
+
+        boolean isUnival = true;
+        if (root.left != null) {
+            isUnival = isUnivalSubtree(root.left) && root.left.val == root.val;
+        }
+        if (root.left != null) {
```
altayhunoglu, line 23:
```suggestion
        if (root.right!= null) {
```
Quite fine when we fix this:
image

src/main/java/algorithms/curated170/medium/countunivaluesubtrees/CountUnivalueSubtreesDFSPassValue.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    private boolean isUnivaluePart(TreeNode root, int parentVal)
+    {
+        if(root == null)
+        {
+            return true;
+        }
+
+        if(!isUnivaluePart(root.left, root.val) & !isUnivaluePart(root.right, root.val))
```
altayhunoglu, line 21:
```suggestion
        if(!isUnivaluePart(root.left, root.val) | !isUnivaluePart(root.right, root.val))
```
If we fix it, it is quite fine:
image
As a not for the future audience:
`A | B`:  First, evaluate A then evaluate B. Finally, apply the OR.

 `A || B`: First, evaluate A, and only if A is false then evaluate B. Finally, apply the OR.

ErdemT09: Same goes for the `&` and `&&` operators. `&&` does not run and check the right side of the statement if the left side is false, because that guarantees that the expression will be false. `&&` and `||` are thus useful for things like null-checks, sanity tests and prioritizing faster operations. `&`, however, also runs the right side of the statement even if the left side is false, i.e. even when the expression is confirmed to be false. In this question, we have to DFS through all the nodes, so we need to use `&` so as to DFS to the right-side even if the left-side is not a valid subtree.

altayhunoglu (approved):
Thank you for the valuable contribution.

PR #291, opened 2021-05-29
325. Maximum Size Subarray Sum Equals k 

ErdemT09:
Resolves: #61 

A maximum Size subarray sum that equals `k` is an equivalent expression to this:
The sum of all numbers up to a certain index `i` equals to some sum `s`.
If we subtract k from s, and the result was previously encountered at some index `j`, that means that we have a subarray of length `i-j` that equals to `k`.
We can illustrate this with an example:
Consider the array `[7, 1, -2, 5, -1, 3]` and the value `k = 3`. The current sum at index `i = 4` is `(7+1+(-2)+5+(-1)) = 10`. When subtract `k` from this current sum, we get 7. We have previously encountered 7 at the index `j = 0`. So, `i-j = 4` will yield us a range of 4. We can calculate ranges as such and compare them to each other easily with a maximum operator and a store them in a Map.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/MaximumSizeSubarrayEqualsK.java
+++ b/src/main/java/algorithms/curated170/medium/MaximumSizeSubarrayEqualsK.java
@@ -1,25 +1,40 @@
 package algorithms.curated170.medium;
 
-import java.util.HashMap;
+import java.util.HashMap;    
+
 
 public class MaximumSizeSubarrayEqualsK {
+
       public int maxSubArrayLen(int[] nums, int k) {
-            HashMap<Integer, Integer> indexStart = new HashMap<>();
-            int tempSum = 0;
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Code is quite fine with 29ms.
But this problem needs some kind of visualization, mostly similar to a tabulatioın problem.
It would be great if we add this when we are available.

altayhunoglu:
In order to visualize and elaborate your wonderful explanation a bit more:

We can classify this question as a `Range Query Problem` in which we query a **prefix array** (that contains the cumulative sum of the initial array)

Thinking in that way, make it easier to reduce the problem like this:

"Find the `maximum distance` of `any two indices in the prefix array` of whose values sum up to `k`"

IMG-1643

Issue #185, opened 2021-05-09
1101. The Earliest Moment When Everyone Become Friends


altayhunoglu:
Varsayılan kod:
```
class Solution {
    public int earliestAcq(int[][] logs, int N) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/117588016-83539300-b129-11eb-8aef-79b90fb4b03c.png)
![image](https://user-images.githubusercontent.com/60903744/117588021-8babce00-b129-11eb-9e24-01fbf29ffb12.png)
![image](https://user-images.githubusercontent.com/60903744/117588026-94040900-b129-11eb-9e06-68b3e19bac41.png)

altayhunoglu:
Solved: https://github.com/spiralgo/algorithms/issues/31

ErdemT09:
Solved:
https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java

PR #292, opened 2021-06-01
Adds Union-find solution with PQ  (1101. The Earliest Moment When Everyone Become Friends)

altayhunoglu:
Resolves:
https://github.com/spiralgo/algorithms/issues/185

Question asks:
"**Return the earliest time for which every person became acquainted with every other person.**"

It is a textbook Union-Find question.

Therefore we can **paraphrase** the question in terms of Union-Find like this:
"`Return the earliest time for which all ids (of people) became member of the same (union, one) group.`"

We can also **rephrase** above Union-find statement like this:
"`Return the earliest time for which all ids (of people) became children of the same parent.`"

 Logs are not sorted by date.
 We can use either **PQ** or **Arrays.sort**. 
 Both have almost the same efficiency.
 
 We need to create a Log class to represent any log object, similar to what we have done in https://github.com/spiralgo/algorithms/pull/286 for Point.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java
+++ b/src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java
@@ -49,7 +49,9 @@ public class TheEarliestMomentWhenEveryoneBecomeFriends {
 
         int findParent(int id) {
             while (parent[id] != id) {
-                id = parent[id];
+                int prevParent = parent[id];
+                parent[id] = parent[prevParent];
+                id = prevParent;
             }
             return id;
         }
--- a/src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java
+++ b/src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java
@@ -1,0 +1,70 @@
+package algorithms.curated170.medium;
+
+import java.util.PriorityQueue;
+
+public class TheEarliestMomentWhenEveryoneBecomeFriends {
+
+    class Solution {
+
+        int[] parent;
+        int[] child;
+
+        public int earliestAcq(int[][] logs, int n) {
+            parent = new int[n];
+            child = new int[n];
+
+            int id_A, id_B, parent_A, parent_B;
+
+            for (int i = 0; i < n; i++) {
+                parent[i] = i;
+                child[i] = 1;
+            }
+
+            PriorityQueue<Log> pq = new PriorityQueue<>((a, b) -> a.timestamp - b.timestamp);
+            for (int i = 0; i < logs.length; i++) {
+                pq.add(new Log(logs[i][0], logs[i][1], logs[i][2]));
+            }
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java
```diff
@@ -0,0 +1,70 @@
...
+
+            return -1;
+        }
+
+        int findParent(int id) {
+            while (parent[id] != id) {
+                id = parent[id];
+            }
+            return id;
+        }
```
ErdemT09, line 55: Is it possible to compress parents?
```suggestion
        int findParent(int id) {
            while (parent[id] != id) {
                parent[id] = parent[parent[id]];
                id = parent[id];
            }
            return id;
        }
```

altayhunoglu: I have to check up on it again, but it looks that the first code already compresses the tree from the beginning.

ErdemT09: It shouldn't on the grounds that the parent of the children of a parent remain same when the parent of the parent is changed.
For example the pairs:     [ 1-0, 2-1, 3-2 ]  would result in the tree ( 3->2->1->0 (root-parent) )

ErdemT09: It shouldn't on the grounds that the parent of the children of a parent remain same when the parent of the parent is changed.
For example the pairs: [ 1-0, 2-1, 3-2, 0-4, 4-5, 5-6 ] would result in the tree ( (1,2,3)->0->4->5->6 (root-parent) )

ErdemT09: Compression in and of itself has a cost but it should render more efficient if such IDs are called perhaps hundreds of times.

altayhunoglu: Oke, we accept the suggestion. Thanks for the feedback.

ErdemT09:
Are there any performance differences between the two?

src/main/java/algorithms/curated170/medium/TheEarliestMomentWhenEveryoneBecomeFriends.java
altayhunoglu: Compression takes a little bit more.
But it is up to the data population used by test cases in LeetCode.

image
image

altayhunoglu:
> Are there any performance differences between the two?

Compression takes a little bit more.
But it is up to the data population used by test cases in LeetCode.

image
image

ErdemT09:
There might also be a DFS solution. I personally think we can skip that approach for now, as we have done relatively enough DFS questions by now.

ErdemT09 (approved):

Issue #237, opened 2021-05-15
158. Read N Characters Given Read4 II - Call multiple times


altayhunoglu:
Varsayılan kod:

```
/**
 * The read4 API is defined in the parent class Reader4.
 *     int read4(char[] buf4); 
 */

public class Solution extends Reader4 {
    /**
     * @param buf Destination buffer
     * @param n   Number of characters to read
     * @return    The number of actual characters read
     */
    public int read(char[] buf, int n) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380881-1be88800-b5ee-11eb-8f1e-f0ea38168708.png)
![image](https://user-images.githubusercontent.com/60903744/118380885-2571f000-b5ee-11eb-8b96-7a3a4081b72d.png)
![image](https://user-images.githubusercontent.com/60903744/118380888-2c98fe00-b5ee-11eb-998f-66660267b1ce.png)
![image](https://user-images.githubusercontent.com/60903744/118380892-328edf00-b5ee-11eb-8b5f-6e9ec1d0f9ca.png)
![image](https://user-images.githubusercontent.com/60903744/118380894-40446480-b5ee-11eb-9cca-831657b43f79.png)
![image](https://user-images.githubusercontent.com/60903744/118380896-45a1af00-b5ee-11eb-8856-70e5b4404161.png)

PR #293, opened 2021-06-01
Read4

altayhunoglu:
Updates: https://github.com/spiralgo/algorithms/issues/161, resolves: https://github.com/spiralgo/algorithms/issues/237

The only difference between the easier one and the harder one:

The harder one (https://github.com/spiralgo/algorithms/issues/237) needs to keep the states of `index`, `buf`, `length`.

Therefore we had to define them as global variables. 

But in the easier one (https://github.com/spiralgo/algorithms/issues/161), we can use them as local variables.

 Suppose, we have a file on a remote server. The content of the file might be "**TheSpiralWillExpand**"
 As we do not have direct access to the file, we need to use the Read4 interface to read the content.
 
IMG-1644

### DIFF
--- a/src/main/java/algorithms/curated170/easy/ReadNCharactersGivenRead4.java
+++ b/src/main/java/algorithms/curated170/easy/ReadNCharactersGivenRead4.java
@@ -22,6 +22,21 @@ public class ReadNCharactersGivenRead4 extends Reader4 {
      
         return copiedChars;
     }
+    
+    public int read2(char[] buf, int n) {
+        int copiedChars = 0;
+     
+        while(copiedChars < n){
+            int length = read4(buf4);
+            for(int i=0; i<length && copiedChars < n; i++){
+                buf[copiedChars++] = buf4[i];
+            }
+          
+            if(length < 4) break;
+        }
+
+        return copiedChars;
+    }
     public static void main(String[] args) {
         ReadNCharactersGivenRead4 readNCharactersGivenRead4 = new ReadNCharactersGivenRead4();
         char[] buf = new char[19];
--- a/src/main/java/algorithms/curated170/easy/ReadNCharactersGivenRead4.java
+++ b/src/main/java/algorithms/curated170/easy/ReadNCharactersGivenRead4.java
@@ -1,36 +1,33 @@
 package algorithms.curated170.easy;
-class Reader4{
-      int read4(char[] buf){
-          
-          //TODO
-          return -1;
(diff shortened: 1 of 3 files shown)


src/main/java/algorithms/curated170/easy/Reader4.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class Reader4 {
+
+    String fileData = "TheSpiralWillExpand";
+ 
+    public static int start = 0;
+    public int read4(char[] buf) {
+
+     
+
```
altayhunoglu, line 11:
```suggestion
```

src/main/java/algorithms/curated170/easy/ReadNCharactersGivenRead4.java
```diff
@@ -1,36 +1,32 @@
...
+       
+            if (index == length) {
+                length = read4(buf4);
+                index = 0;
+                if (length == 0) {
+                    break;
+                }
+            }
+            
+            buf[copiedChars++] = buf4[index++];
```
ErdemT09, line 20: read4 should read 4 chars at a time but we are only incrementing by 1 in our destination char array.  Why don't we just call it then loop through the 4 or less elements of buf4 to assign them to buf, instead of creating a gap between length and index?

ErdemT09: Similar to the last solution to some degree.

altayhunoglu: Could you share the code in your mind?

ErdemT09:
```java
    public int read(char[] buf, int n) {
        int copiedChars = 0;
        char[] buf4 = new char[4];
        while(copiedChars< n){
            int length = read4(buf4);
            for(int i=0; i<length && copiedChars < n; i++){
                buf[copiedChars++] = r4[i];
            }
            if(length < 4) break;
        }

        return copiedChars;
    }
```

altayhunoglu: Oke. This is also added as the second approach.

ErdemT09 (approved):
Pretty substandard for a hard question. Nevertheless, the code is well-structured. Thank you for your efforts.

PR #294, opened 2021-06-03
254. Factor Combinations

ErdemT09:
Resolves: #60 

### Algorithm
 In this question, there are few thing we aren't allowed to do:
1. The factors don't include the pair `{input number, 1}`
2. We don't include the duplicate pairs, for example `{3,9}` and `{9,3}` by the input 27

For this, we utilize a algorithm similar to that in #43. We keep our factors in a stack. We start iterating through our numbers initially at 2, then at the number that divides the other number when we reach the loop inside the method. At 24, 2 divides it. So we send a new backtracking branch with the values 12 and 2, we add the divisor 2 to the stack. Initially, to get around the 1. issue, we don't add the input number (the one other than the divisor) when the stack is empty, which could only occur when we initially call the method or the initial number is not factorizable as expected in the problem statement. Otherwise, we add the input number to the stack and add the stack to our factor combinations, then we pop the input number from the stack. Recursively, we also iterate through the possible numbers here and check if can branch off from there too.
We only iterate through possible divisors up to `i ≤ n/i`, because otherwise we would encounter the same branch again. (For example `12/3` and `12/4` complement and disqualify themselves as such). 
Eventually, after the backtracking branch is complete, we remove the divisor from the stack to empty up place for the new possible divisor branches.

Here is video illustrating how this algorithm works for the input `n=12`: https://youtu.be/ws8qX__wxq4

### Algorithm
This algorithm is actually quite simple and basically works the same way as the prime factorization one would have done in their school years. We take the number, we check if it is divisible by a prime starting with 2. If not, we check if it is divisible by the next prime 3 and then 5, 7, etc. The point is that the minimum prime of a number can at most be its square root, think of 121 as an example. So, we can apply the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) here, as we had done with #94. Up to this limiting maximum prime, we "sieve" through the factors of current primes and eliminate these. The remaining numbers are prime, so we add them to our list of primes and check whether they are our next prime factor.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/FactorCombinations.java
+++ b/src/main/java/algorithms/curated170/medium/FactorCombinations.java
@@ -6,36 +6,36 @@ import java.util.Stack;
 
 public class FactorCombinations {
 
-    List<List<Integer>> res;
-    Stack<Integer> temp;
+    List<List<Integer>> combinations;
+    Stack<Integer> currFactors;
 
     void backtracking(int n, int index) {
         if (n > 1) {
-            if (!temp.isEmpty()) {
-                temp.add(n);
-                res.add(new ArrayList<>(temp));
-                temp.pop();
+            if (!currFactors.isEmpty()) {
+                currFactors.add(n);
+                combinations.add(new ArrayList<>(currFactors));
+                currFactors.pop();
             }
         }
 
         for (int i = index; i <= n / i; i++) {
             if (n % i == 0) {
-                temp.add(i);
+                currFactors.add(i);
                 backtracking(n / i, i);
-                temp.pop();
+                currFactors.pop();
             }
         }
     }
 
     public List<List<Integer>> getFactors(int n) {
-        res = new ArrayList<>();
+        combinations = new ArrayList<>();
         if (n < 4)
         {
-            return res;
+            return combinations;
         }
-        temp = new Stack<>();
+        currFactors = new Stack<>();
         backtracking(n, 2);
-        return res;
+        return combinations;
     }
 
     public static void main(String[] args) {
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/medium/FactorCombinations.java
```diff
@@ -0,0 +1,47 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Stack;
+
+public class FactorCombinations {
```
altayhunoglu, line 7: image

src/main/java/algorithms/util/math/PrimeFactorsGenerator.java
```diff
@@ -0,0 +1,117 @@
+package algorithms.util.math;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class PrimeFactorsGenerator {
```
altayhunoglu, line 8: image

ErdemT09: This is not a solution to the problem. This is just an algorithm for generating prime factors of a number. That's why it is in the math package.

altayhunoglu: Oke, now I understand. I will review then.

altayhunoglu:
Your new debug-like explanation technique is really good by the way.

https://youtu.be/ws8qX__wxq4

altayhunoglu (approved):
Nice codes and I have nothing to add about the coding side.
We can draw a backtracking tree for the code when we are available.

ErdemT09:
> We can draw a backtracking tree for the code when we are available.

I tried to illustrate the method calls like this. 
image

altayhunoglu (approved):
Thanks for the valuable contribution Erdem.

altayhunoglu:
Thanks for the visualization by the way. It is a nice way to demonstrate the calls.

PR #295, opened 2021-06-04
1182. Shortest Distance to Target Color 

ErdemT09:
Resolves: #46 
### Algorithm:
Instead of looking up linearly where the closest target color is, at the beginning, we can add the indices of each color to an array list. While handling the query, we can just binary search through the index list of the target color to find the closest index. The binary search that we use here is a bit modified. If the given index is larger than the last value or smaller than the first value, we must return the respective value, we need to do an initial check for this. Otherwise, we perform an ordinary binary search and compare the values at left and right at the end and return the closer one.

### Algorithm:
At any index, the closest of each color is the minimum of the closest left and the closest right. This can be implemented in linear time by first looping through the colors array incrementally, then decrementally. After that, we have an array as a lookup table for all possible colors and indices that the query may offer.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/shortestdistancetotargetcolor/ShortestDistanceToTargetColorPreprocessing.java
+++ b/src/main/java/algorithms/curated170/medium/shortestdistancetotargetcolor/ShortestDistanceToTargetColorPreprocessing.java
@@ -39,7 +39,10 @@ public class ShortestDistanceToTargetColorPreprocessing {
     private void compareClosestRightColors(int[] colors) {
         for (int i = colors.length - 2; i >= 0; i--) {
             for (int c = 1; c <= 3; c++) {
-                closestDistTable[i][c] = Math.min(closestDistTable[i][c], closestDistTable[i + 1][c] + 1);
+                int closestRight = closestDistTable[i + 1][c];
+                if (closestRight != NOT_FOUND) {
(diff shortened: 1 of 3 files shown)


src/main/java/algorithms/curated170/medium/shortestdistancetotargetcolor/ShortestDistanceToTargetColorPreprocessing.java
```diff
@@ -0,0 +1,72 @@
...
+
+        assignClosestLeftColors(colors);
+
+        compareClosestRightColors(colors);
+
+        return handleQueriesLookup(queries);
+    }
+
+    private void assignClosestLeftColors(int[] colors) {
+        closestDistTable = new int[colors.length][];
```
altayhunoglu, line 22: What does the second dimension of closestDistTable store?

ErdemT09: Closest index of each color.

altayhunoglu (approved):

ErdemT09:
How does the new improved solution run?

altayhunoglu:
> How does the new improved solution run?

Yeah, it is faster now:
image

PR #296, opened 2021-06-04
Adds PaintFence DP solution and unit test. (276 Paint Fence)

altayhunoglu:
Resolves:
https://github.com/spiralgo/algorithms/issues/81

As we can use the same conclusion and previous data for n>2, it is a textbook DP problem.
I tried to demystify it by visualization.

IMG-1651


ErdemT09:
I had previously checked this question, but the image actually explains the algorithm without leaving a need to look up the code. The code itself is pretty lean and well. Approved.

ErdemT09 (approved):

PR #297, opened 2021-06-05
121. Best Time to Buy and Sell Stock

ErdemT09:
Resolves: https://leetcode.com/problems/best-time-to-buy-and-sell-stock/

In this question, we have to return the maximum profit we earn by buying some stock and selling it later. As there is a possibility that we may never earn anything, a `maxProfit` must be initialized with value 0. At every stock price, we naturally make 2 decisions:
1. Is this cheaper than the previous cheapest stock?
2. Is selling this (price of this - last cheapest price) more profitable than the previous maximum profit?

For this, we additionally store the last cheapest price in some variable `currBuy`. This leads to code structure where we just linearly iterate through prices and do these 2 comparisons using basic min and max methods.

Logically, if we update the cheapest price at any stock, that means that trying to sell at that stock would result in 0 profit. We can ignore calculating the profit in such cases.

### DIFF
--- a/src/main/java/algorithms/easy/BestTimeToBuyAndSellStock.java
+++ b/src/main/java/algorithms/easy/BestTimeToBuyAndSellStock.java
@@ -3,12 +3,12 @@ package algorithms.easy;
 public class BestTimeToBuyAndSellStock {
 
 	public int maxProfit(int[] prices) {
-		int buy = Integer.MAX_VALUE, sell = 0;
+		int currBuy = Integer.MAX_VALUE, maxProfit = 0;
 		for (int p : prices) {
-			buy = Math.min(p, buy);
-			sell = Math.max(sell, p - buy);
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/easy/BestTimeToBuyAndSellStock.java
```diff
@@ -0,0 +1,24 @@
+package algorithms.easy;
+
+public class BestTimeToBuyAndSellStock {
+
+	public int maxProfit(int[] prices) {
+		int currBuy = Integer.MAX_VALUE, maxProfit = 0;
+		for (int p : prices) {
+			if (p > currBuy) {
```
altayhunoglu, line 8:
```suggestion
			if (p < currBuy) {
```
When we fix this:

image

altayhunoglu (approved):

Issue #227, opened 2021-05-15
265. Paint House II


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int minCostII(int[][] costs) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380675-37eb2a00-b5ec-11eb-9c61-ca4eec6228b6.png)
![image](https://user-images.githubusercontent.com/60903744/118380677-3d487480-b5ec-11eb-904a-c354b680a9ab.png)
![image](https://user-images.githubusercontent.com/60903744/118380680-420d2880-b5ec-11eb-9f8d-211f50ef6cd5.png)
![image](https://user-images.githubusercontent.com/60903744/118380683-48030980-b5ec-11eb-9862-fd0ddc72dd3c.png)

PR #298, opened 2021-06-05
Adds DP solution to 265. Paint House II

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/227

It is a generalized variation of https://github.com/spiralgo/algorithms/issues/44.

At this time, the number of colors is `k` instead of `3`.

Therefore, we need to find a generalized statement to express the intuition for both questions.

For https://github.com/spiralgo/algorithms/issues/44:

  ```java 
  costs[i][0] += Math.min(costs[i-1][1], costs[i-1][2]);
 ```

 We are accumulating the cost of `the second previous minimum` for an index here.
 We use `"the second"` because we do not want "`two adjacent houses have the same color`".
 
 When we had `3` alternative colors, we could utilize "Math.min" to decide the second previous minimum.
  But when we have `k` alternative colors, we have to iterate through the colors to keep the state of:
 `preMin ` (previous minimum),  `preSecondMin ` (previous second minimum) , `preMinIdx ` (the index of previous color with the minimum cost)

### DIFF
--- a/src/main/java/algorithms/curated170/hard/PaintHouseII.java
+++ b/src/main/java/algorithms/curated170/hard/PaintHouseII.java
@@ -1,0 +1,36 @@
+package algorithms.curated170.hard;
+
+public class PaintHouseII {
+
+    public int minCostII(int[][] costs) {
+        int houses = costs.length;
+        int colors = costs[0].length;
+
+        int preMin = 0, preSecondMin = 0, preMinIdx = -1;
+        
+        for (int i = 0; i < houses; i++) {
+            int curMin = Integer.MAX_VALUE, curSecondMin = Integer.MAX_VALUE, curMinIdx = -1;
+            for (int j = 0; j < colors; j++) {
+                
+                int val = costs[i][j] + (j == preMinIdx ? preSecondMin : preMin);
+                
+                if (curMinIdx == -1) {
+                    curMin = val;
+                    curMinIdx = j;
+                } else if (val < curMin) {
+                    curSecondMin = curMin;
+                    curMin = val;
+                    curMinIdx = j;
+                } else if (val < curSecondMin) {
+                    curSecondMin = val;
+                }
+            }
+           
+            preMin = curMin;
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/hard/PaintHouseII.java
```diff
@@ -0,0 +1,36 @@
...
+                    curSecondMin = val;
+                }
+            }
+           
+            preMin = curMin;
+            preSecondMin = curSecondMin;
+            preMinIdx = curMinIdx;
+        }
+        return preMin;
+    }
```
ErdemT09, line 34: I think its better to use a for-each loop for the houses.
```suggestion
    public int minCostII(int[][] costs) {
        int colors = costs[0].length;

        int preMin = 0, preSecondMin = 0, preMinIdx = -1;

        for (int[] house : costs) {
            int curMin = Integer.MAX_VALUE, curSecondMin = Integer.MAX_VALUE, curMinIdx = -1;
            for (int j = 0; j < colors; j++) {

                int val = house[j] + (j == preMinIdx ? preSecondMin : preMin);

                if (curMinIdx == -1) {
                    curMin = val;
                    curMinIdx = j;
                } else if (val < curMin) {
                    curSecondMin = curMin;
                    curMin = val;
                    curMinIdx = j;
                } else if (val < curSecondMin) {
                    curSecondMin = val;
                }
            }

            preMin = curMin;
            preSecondMin = curSecondMin;
            preMinIdx = curMinIdx;
        }
        return preMin;
    }
```

altayhunoglu: Absolutely. It is both faster and more readable.
image

ErdemT09 (approved):
The code is pretty great at expressing itself. As a naive solution, I first though of iterating through each previous house color as we had done in the prequel, but this is more clever, sort of like a double dynamic programming approach.

Thank you for contributing an algorithm to the repository for this day. I will add one too.
Approved.

Issue #218, opened 2021-05-15
272. Closest Binary Search Tree Value II


altayhunoglu:
https://leetcode.com/problems/closest-binary-search-tree-value-ii/

Varsayılan kod:

```
/**
 * Definition for a binary tree node.
 * public class TreeNode {
 *     int val;
 *     TreeNode left;
 *     TreeNode right;
 *     TreeNode() {}
 *     TreeNode(int val) { this.val = val; }
 *     TreeNode(int val, TreeNode left, TreeNode right) {
 *         this.val = val;
 *         this.left = left;
 *         this.right = right;
 *     }
 * }
 */
class Solution {
    public List<Integer> closestKValues(TreeNode root, double target, int k) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380487-c199f800-b5ea-11eb-8f2c-486d1d6b9452.png)
![image](https://user-images.githubusercontent.com/60903744/118380490-c9f23300-b5ea-11eb-873a-9d24a1d864f6.png)
![image](https://user-images.githubusercontent.com/60903744/118380492-cfe81400-b5ea-11eb-8948-39bdafcb1839.png)

PR #299, opened 2021-06-05
Adds Inorder Traversal solution with LinkedList (272. Closest Binary Search Tree Value II)

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/218

Even if it is a variant of https://github.com/spiralgo/algorithms/issues/95, just because this question asks for a list of the closest value; using Binary Search seems inefficient for the question.

There are several solutions using PQ, Array Sort, or an algorithm called QuickSelect which is by Tony Hoare.

But, the LinkedList solution is quite straightforward. We can add the others if we need them in the future.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/ClosestBinarySearchTreeValueIILinkedList.java
+++ b/src/main/java/algorithms/curated170/hard/ClosestBinarySearchTreeValueIILinkedList.java
@@ -1,0 +1,39 @@
+package algorithms.curated170.hard;
+
+import algorithms.datastructures.TreeNode;
+import java.util.LinkedList;
+import java.util.List;
+
+public class ClosestBinarySearchTreeValueIILinkedList {
+
+    LinkedList<Integer> result;
+
+    public List<Integer> closestKValues(TreeNode root, double target, int k) {
+        result = new LinkedList<>();
+
+        if (root == null) {
+            return result;
+        }
+
+        inorder(root, target, k);
+
+        return result;
+    }
+
+    private void inorder(TreeNode curr, double target, int k) {
+        if (curr == null) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/hard/PaintHouseII.java
```diff
@@ -0,0 +1,36 @@
+package algorithms.curated170.hard;
+
+public class PaintHouseII {
```
ErdemT09, line 3: This also came along.

ErdemT09:
I agree, this solution is pretty clear.

src/main/java/algorithms/curated170/hard/PaintHouseII.java
altayhunoglu: Sorry for that. It will not affect but disturbing.

src/main/java/algorithms/curated170/hard/ClosestBinarySearchTreeValueIILinkedList.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private void inorder(TreeNode curr, double target, int k) {
+        if (curr == null) {
+            return;
+        }
+
+        inorder(curr.left, target, k);
+        if (result.size() < k) {
+            result.add(curr.val);
+        } else if (Math.abs(curr.val - target) < Math.abs(result.getFirst() - target)) {
```
ErdemT09, line 31: As a note, we are comparing the current with the first because the first element is most probably the furthest from our target value. If it's not so, that means that the rest of the elements are further around some pivot inside the list, so we only get further away from this pivot when we do inorder travesal.

ErdemT09 (approved):
Approved.

PR #300, opened 2021-06-05
298. Binary Tree Longest Consecutive Sequence

ErdemT09:
Resolves: #58 

## Algorithm:
The intuition is pretty simple. If a node has a consecutive sequence (hereinafter cs) with the left branch and another on the right, the length of the cs at this node would have been the maximum of these two cs. If there is no consecutiveness with either left or right, we can assume than the length of the cs incoming from that direction is 0. If both are 0, the length of cs at this node can remain 1. 
We can think this backwards too. If a node's value is consecutive to its parent's, then it should return the length of its cs, otherwise 0. At the base case of null leaves, they should return a length of 0. For this, we send the parent's value to a DFS of both left and right branches, receive their values and add the maximum one to 1. If the length of cs here is longer than the global maximum, we update it. I think the traversal here can be described as "postorder".

### DIFF
--- a/src/main/java/algorithms/curated170/medium/BinaryTreeLongestConsecutiveSequence.java
+++ b/src/main/java/algorithms/curated170/medium/BinaryTreeLongestConsecutiveSequence.java
@@ -1,0 +1,48 @@
+package algorithms.curated170.medium;
+
+import algorithms.datastructures.TreeNode;
+
+public class BinaryTreeLongestConsecutiveSequence {
+
+     private int maxLength = 0;
+
+    public int longestConsecutive(TreeNode root) {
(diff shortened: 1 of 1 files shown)


altayhunoglu:
> ## 
>  I think the traversal here can be described as "postorder".

You are right. 
At in-order traversal nodes are like apples falling down the tree.

At post-order traversal, it is like wrapping a tree in a veil or blanket that touches just the tree's outer nodes.

Therefore post-order looks appropriate to calculate that kind of distances.


 image

altayhunoglu (approved):
Quite fine Erdem. Thanks.
image

PR #301, opened 2021-06-06
651. 4 Keys Keyboard

ErdemT09:
Resolves: #47 

The thing that we have to notice is that select all and copy operations are always done together. After having copied some 'A's, it is always (except when we have copied just one character, then it is just equal) more profitable to paste than to insert new single characters. After copying:

- pasting 1 time would mean doubling the count (cost = 3)
- pasting 2 times would mean tripling the count (cost = 4)
- pasting 3 times would mean quadrupling the count (cost = 5)
- pasting 4 times would mean quintupling the count (cost = 6)
- pasting W-2 times would mean multiplying the count by W-1 (cost = W)

Multiplying by 6 is never more profitable than multiplying by 3 and then 2 (cost 7 vs 7), or by 7  never more profitable than multiplying by 4 and then 2 (cost 8 vs 8, but actually 1 more character gain). Referring to the image, multiplying by any number greater than 5 (N*2 = 6) is never profitable.
image
Multiplying by 2 two times is never more profitable than multiplying by multiplying by four. 
That is why only amount of multiplications that we should consider are 3, 4 and 5.

At some presses P, we can always go back to 

- the best of 6 presses ago and multiply it by 5 and compare
- the best of 5 presses ago and multiply it by 4 and compare
- the best of 4 presses ago and multiply it by 3 and compare

In this way, our algorithm stores the best of previous number of presses and compares them with the current. 

Eventually, 4 must be the most optimal choice:
For the same cost of 20, we can either multiply by a = 3^5 or a < 4^4
For the same cost of 30, we can either multiply by a = 5^5 or a < 4^6

Therefore, we can state that we will always be multiplying by 4 from some point on. We can calculate that point using the dynamic programming algorithm and just have it some lookup array. If the given number of presses is larger than the size of that array, we just look up its corresponding value 5*A ago and multiply that looked up value by 4^A.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
+++ b/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.medium;
+package algorithms.curated170.medium.fourkeyskeyboard;
 
 public class FourKeysKeyboardDP {
 
--- a/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
+++ b/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
@@ -1,0 +1,28 @@
+package algorithms.curated170.medium;
+
+public class FourKeysKeyboardDP {
+
+    public int maxA(int N) {
+        int[] best = new int[N + 1];
+
+        for (int press = 1; press <= N; press++) {
+            best[press] = best[press - 1] + 1;
+            if (press >= 6) {
+                for (int x = press - 6; x < press - 3; x++) {
+                    int pasteTimes = press - 1 - x;
+                    best[press] = Math.max(best[press], best[x] * pasteTimes);
+                }
+            }
+        }
+        return best[N];
+    }
+
+    public static void main(String[] args) {
+        var solution = new FourKeysKeyboardDP();
+        
+        for (int i = 0; i < 50; i++) {
+            
+        }
+
+    }
+}
--- a/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
+++ b/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
@@ -8,10 +8,9 @@ public class FourKeysKeyboardDP {
         for (int press = 1; press <= N; press++) {
             best[press] = best[press - 1] + 1;
             if (press >= 6) {
-                for (int x = press - 6; x < press - 3; x++) {
-                    int pasteTimes = press - 1 - x;
-                    best[press] = Math.max(best[press], best[x] * pasteTimes);
-                }
+                best[press] = Math.max(best[press], best[press - 6] * 5);
+                best[press] = Math.max(best[press], best[press - 5] * 4);
+                best[press] = Math.max(best[press], best[press - 4] * 3);
             }
         }
         return best[N];
@@ -19,9 +18,9 @@ public class FourKeysKeyboardDP {
 
     public static void main(String[] args) {
         var solution = new FourKeysKeyboardDP();
-        
+
         for (int i = 0; i < 50; i++) {
-            
+            System.out.println(solution.maxA(i));
         }
 
     }
--- a/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
+++ b/src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
@@ -7,7 +7,7 @@ public class FourKeysKeyboardDP {
 
         for (int press = 1; press <= N; press++) {
             best[press] = best[press - 1] + 1;
-            if (press >= 6) {
+            if (press > 6) {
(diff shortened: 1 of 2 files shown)


altayhunoglu:
Then, the reason that "3" (AAA) is the basis for the solution should be:  `Ctrl-A + Ctrl-C + Ctrl-V` can be done in at least 3 steps.
 If one had just "n" unit currency, then he would spend it to insert just three single chars instead of using `Ctrl-A + Ctrl-C + Ctrl-V`

ErdemT09:
> Then, the reason that "3" (AAA) is the basis for the solution is: `Ctrl-A + Ctrl-C + Ctrl-V` can be done in at least 3 steps.
> If one had just "n=3" unit currency, then he would spend it to insert just three single chars instead of using `Ctrl-A + Ctrl-C + Ctrl-V`

True, until 7, we never do anything more than incrementing.

altayhunoglu:
> > Then, the reason that "3" (AAA) is the basis for the solution is: `Ctrl-A + Ctrl-C + Ctrl-V` can be done in at least 3 steps.
> 
> > If one had just "n=3" unit currency, then he would spend it to insert just three single chars instead of using `Ctrl-A + Ctrl-C + Ctrl-V`
> 
> 
> 
> True, until 7, we don't do anything more than incrementing.
> 
> 

Then can we claim:
**If n<7, f(n)=n**

altayhunoglu:
> > Then, the reason that "3" (AAA) is the basis for the solution is: `Ctrl-A + Ctrl-C + Ctrl-V` can be done in at least 3 steps.
> > If one had just "n=3" unit currency, then he would spend it to insert just three single chars instead of using `Ctrl-A + Ctrl-C + Ctrl-V`
> 
> True, until 7, we don't do anything more than incrementing.

Because when it is 6:
```

1- PRESS A three times.
2- `Ctrl-A + Ctrl-C + Ctrl-V` 
```

src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
```diff
@@ -0,0 +1,27 @@
+package algorithms.curated170.medium.fourkeyskeyboard;
+
+public class FourKeysKeyboardDP {
```
altayhunoglu, line 3: image

src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public int maxA(int N) {
+        int[] best = new int[N + 1];
+
+        for (int press = 1; press <= N; press++) {
+            best[press] = best[press - 1] + 1;
+            if (press > 6) {
+                best[press] = Math.max(best[press], best[press - 6] * 5);
+                best[press] = Math.max(best[press], best[press - 5] * 4);
+                best[press] = Math.max(best[press], best[press - 4] * 3);
```
altayhunoglu, line 13: You can choose:
```suggestion
               best[press] = Math.max(best[press], best[press - 3] * 2);
                best[press] = Math.max(best[press], best[press - 4] * 3);
                best[press] = Math.max(best[press], best[press - 5] * 4);
```

It is more relatable for me. Because it represents the selection between those choices more cleaner:
after f[i - 3], by ACP total length : f[i - 3] * 2
after f[i - 4], by ACP + P total length : f[i - 4] * 3
after f[i - 5], by ACP + PP total length : f[i - 5] * 4

src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardMath.java
```diff
@@ -0,0 +1,23 @@
+package algorithms.curated170.medium.fourkeyskeyboard;
+
+public class FourKeysKeyboardMath {
```
altayhunoglu, line 3: image

src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
ErdemT09: It should be a multiplication by 3, 4 and 5. Otherwise, I'm changing then.

ErdemT09: Added.

altayhunoglu: It should also work like that:
```java
               best[press] = Math.max(best[press], best[press - 3] * 2);
                best[press] = Math.max(best[press], best[press - 4] * 3);
                best[press] = Math.max(best[press], best[press - 5] * 4);
```

But no problem. This is fine. I approve.

altayhunoglu (approved):
Thanks for your valuable contribution.

src/main/java/algorithms/curated170/medium/fourkeyskeyboard/FourKeysKeyboardDP.java
ErdemT09: Perhaps we never multiply by 5. But theoretically, we may be.

altayhunoglu: Maybe there is a test-case we need. But according to LeetCode, we don't need it.

image

PR #302, opened 2021-06-06
259. 3Sum Smaller

### LINKED ISSUE #55: 259. 3Sum Smaller
https://leetcode.com/problems/3sum-smaller

Bir integer arrayi (int[] nums) ve bir hedef değer (int target) veriliyor.

Şu şartları sağlayacak üçlülerin sayısı isteniyor:
i, j, k bu üçlünün indeksleri olsun:  (0 <= i < j < k < n  olmak üzere)
nums[i] + nums[j] + nums[k] < target şartına uygun üçlüleri listeleyeceğiz.

(Burada n, nums içindeki eleman sayısı. Bununla birlikte toplama elemanlarının yeri değişebilir (commutative) olduğundan, buradaki 0 <= i < j < k < n ifâdesinin tek maksadı, muhtemelen, aynı elemanı iki kez toplamayı engellemek içindir. Yani; i, j, k indekslerinin sırasını değiştirmek toplamayı değiştirmez. Bu bize, orijinal diziyi sıralayabileceğimizi gösterir.)
 
Varsayılan kod:
```
class Solution {
    public int threeSumSmaller(int[] nums, int target) {
        
    }
}
```

image

ErdemT09:
Resolves #55 

### Points to notice:
- Adding 3 numbers being less than some value is equivalent to picking one of these numbers, subtracting it from the target value and the sum of the other 2 being less than target - subtracted value (`target_`). 
- We can then sort the array, loop over it from from the lowest value to the largest, subtract the value from the target and look at the right rest of the array for pairs of suitable numbers.
- We can then start this comparison from the right next of the value. The other compared value should then be the largest value. If some largest value + the left value is smaller than `target_` then all the values smaller then this right large value can be added with the left value to get a sum smaller than `target_`.
- The right pointer must be always larger than the left pointer then.
- If the previous operation is successful, we can increment the left pointer to look for two sums with the number at this pointer. Remember, the array is sorted.
- We should then do this collect the possible two sum combinations each time until the loop condition is false.
- In the main loop, if we can't find any two sums at some index, that means that we won't be able to find two sums for numbers larger than it for further indices.
- We don't always need to initialize the right boundary with the size of the array. Since the numbers are always getting larger, our new first possible right pointer will either be smaller than or equal to the previous first possible. So, we can store that in a variable.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
+++ b/src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
@@ -28,11 +28,15 @@ public class ThreeSumSmallerIterative {
 
         int p1 = start, p2 = rightBoundary;
         int count = 0;
+        boolean foundBounds = false;
         while (p1 < p2) {
             if (nums[p1] + nums[p2] >= target) {
-                rightBoundary--;
+                if (!foundBounds) {
+                    rightBoundary--;
+                }
                 p2--;
             } else {
+                foundBounds = true;
                 count += p2 - p1;
                 p1++;
             }
--- a/src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
+++ b/src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
@@ -1,0 +1,50 @@
+package algorithms.curated170.medium.threesumsmaller;
+
+import java.util.Arrays;
+
+public class ThreeSumSmallerIterative {
+
+    int[] nums;
+    int rightBoundary;
+
+    public int threeSumSmaller(int[] nums, int target) {
+
+        this.nums = nums;
+        int count = 0;
+        rightBoundary = nums.length - 1;
+        Arrays.sort(nums);
+
+        for (int i = 0; i < nums.length - 2; i++) {
+            int countAtIdx = findTwoSumsSmaller(i + 1, target - nums[i]);
+            if (countAtIdx == 0) {
+                break;
+            }
+            count += countAtIdx;
+        }
+        return count;
+    }
+
+    public int findTwoSumsSmaller(int start, int target) {
+
+        int p1 = start, p2 = rightBoundary;
+        int count = 0;
+        while (p1 < p2) {
+            if (nums[p1] + nums[p2] >= target) {
+                rightBoundary--;
+                p2--;
+            } else {
+                count += p2 - p1;
+                p1++;
+            }
+        }
+        return count;
+    }
+
+    public static void main(String[] args) {
+
+        var solution = new ThreeSumSmallerIterative();
+
+        int[] nums = new int[] { 0, 1, 3, 4, 8, 9, 11, 12 };
+        System.out.println(solution.threeSumSmaller(nums, 8));
+    }
+}
--- a/src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
+++ b/src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
@@ -8,7 +8,10 @@ public class ThreeSumSmallerIterative {
     int rightBoundary;
 
     public int threeSumSmaller(int[] nums, int target) {
-
+        if (nums.length < 3) {
+            return 0;
+        }
+        
         this.nums = nums;
         int count = 0;
         rightBoundary = nums.length - 1;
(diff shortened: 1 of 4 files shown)


ErdemT09:
I also try adding unit tests for this problem.

src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
```diff
@@ -0,0 +1,54 @@
+package algorithms.curated170.medium.threesumsmaller;
+
+import java.util.Arrays;
+
+public class ThreeSumSmallerIterative {
```
altayhunoglu, line 5: Great: 
image

src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerRecursion.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Arrays;
+
+public class ThreeSumSmallerRecursion {
+
+    int[] nums;
+    int target;
+    int rightBoundary;
+
+    public int threeSumSmaller(int[] nums, int target) {
+
```
altayhunoglu, line 12: It is also great, after fixing the if check:
image

src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
altayhunoglu: I just copy the explanation to track the code while reading it.
> Resolves #55
> 
> ## Algorithm:
> ### Points to notice:
> * Adding 3 numbers being less than some value is equivalent to picking one of these numbers, subtracting it from the target value and the sum of the other 2 being less than target - subtracted value (`target_`).
> * We can then sort the array, loop over it from from the lowest value to the largest, subtract the value from the target and look at the right rest of the array for pairs of suitable numbers.
> * We can then start this comparison from the right next of the value. The other compared value should then be the largest value. If some largest value + the left value is smaller than `target_` then all the values smaller then this right large value can be added with the left value to get a sum smaller than `target_`.
> * The right pointer must be always larger than the left pointer then.
> * If the previous operation is successful, we can increment the left pointer to look for two sums with the number at this pointer. Remember, the array is sorted.
> * We should then do this collect the possible two sum combinations each time until the loop condition is false.
> * In the main loop, if we can't find any two sums at some index, that means that we won't be able to find two sums for numbers larger than it for further indices.
> * We don't always need to initialize the right boundary with the size of the array. Since the numbers are always getting larger, our new first possible right pointer will either be smaller than or equal to the previous first possible. So, we can store that in a variable.

src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerRecursion.java
```diff
@@ -0,0 +1,51 @@
...
+                }
+                p2--;
+            } else {
+                foundBounds = true;
+                count += p2 - p1;
+                p1++;
+            }
+        }
+        return count == 0 ? 0 : count + reduceTwoSumsSmaller(start + 1);
+    }
```
altayhunoglu, line 42: A little bit of simplification and readability fix.

```suggestion
    public int reduceTwoSumsSmaller(int start) {
        int currTarget = target - nums[start];
        int left = start+1, right = rightBoundary;
        int count = 0;
        while (left < right) {
            if (nums[left] + nums[right] < currTarget) {
                  count += right - left;
                  left++;
            } else {
                  right--;
               
            }
        }
        return count == 0 ? 0 : count + reduceTwoSumsSmaller(start + 1);
    }
```

src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerIterative.java
```diff
@@ -0,0 +1,57 @@
...
+                }
+                p2--;
+            } else {
+                foundBounds = true;
+                count += p2 - p1;
+                p1++;
+            }
+        }
+        return count;
+    }
```
altayhunoglu, line 48: Similar simplification may be done here too.

src/main/java/algorithms/curated170/medium/threesumsmaller/ThreeSumSmallerRecursion.java
ErdemT09: How do we make use of the right boundary in this? I don't think itmis being used.

altayhunoglu: We should keep the initial state of `rightBoundary ` for each recursion. If we decrement it (instead of `right`), it will not work.

altayhunoglu (approved):
Thank you for your valuable contribution and collaboration Erdem.

PR #304, opened 2021-06-07
Adds Backtracking solution: 46. Permutations

altayhunoglu:
(Question: https://leetcode.com/problems/permutations/)

Resolves: https://github.com/spiralgo/algorithms/issues/303

It is an easy but important question as it provides a basis for several other questions including https://github.com/spiralgo/algorithms/issues/83

I tried to introduce a template for usual backtracking solutions below:
IMG-1658

For this specific problem, our `choices` are in an array called `nums`
We use the `swap `strategy for both making a choice and generating a permutation.
After we test the generated permutation, passing it as a parameter to the `backtrack` method, we `unmake the choice` using the `swap` method again. (The second **swap** is the point where we do the actual backtracking.)

f592442d-c23c-4262-ad58-75c12c556ab9
[Image source: LeetCode]

### DIFF
--- a/src/main/java/algorithms/curated170/easy/Permutations.java
+++ b/src/main/java/algorithms/curated170/easy/Permutations.java
@@ -1,0 +1,34 @@
+package algorithms.curated170.easy;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Permutations {
+        public List<List<Integer>> permute(int[] nums) {
+            List<List<Integer>> output = new ArrayList<List<Integer>>();
+            backtrack(0, nums, output);
+            return output;
+        }
+
+        public void backtrack(int choiceIndex, int[] nums, List<List<Integer>> output) {
+            if (choiceIndex == nums.length) {
+                ArrayList<Integer> permutation = new ArrayList<Integer>();
+                for (int num : nums) {
+                    permutation.add(num);
+                }
+                output.add(permutation);
+            }
+
+            for (int i = choiceIndex; i < nums.length; i++) {
+                swap(choiceIndex, i, nums);
+                backtrack(choiceIndex + 1, nums, output);
+                swap(i, choiceIndex, nums);
+            }
+        }
+
+        public void swap(int choiceIndex, int i, int[] nums) {
+            int temp = nums[choiceIndex];
+            nums[choiceIndex] = nums[i];
+            nums[i] = temp;
+        }
+}


src/main/java/algorithms/curated170/easy/Permutations.java
```diff
@@ -0,0 +1,34 @@
+package algorithms.curated170.easy;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Permutations {
+        public List<List<Integer>> permute(int[] nums) {
+            List<List<Integer>> output = new ArrayList<List<Integer>>();
```
altayhunoglu, line 8: Passing output as a parameter instead of using a global variable looks more appropriate for "the Law of Demeter"

The class itself does not need to know about output at all.

Global variables are also dangerous in more complex projects as the state of a global variable can be changed by an unrelated method or class accidentally.

In short, it is easier to isolate a variable and track its state using it as a parameter.

Similar arguments against global variables can be found:

https://stackoverflow.com/questions/2216340/the-advantage-disadvantage-between-global-variables-and-function-parameters-in

ErdemT09: They do make the code simpler in small systems like this though. More "functional" approaches might be more suitable for the reliability of complex systems but I think such object oriented approaches okay for here.

src/main/java/algorithms/curated170/easy/Permutations.java
```diff
@@ -0,0 +1,34 @@
...
+            backtrack(0, nums, output);
+            return output;
+        }
+
+        public void backtrack(int choiceIndex, int[] nums, List<List<Integer>> output) {
+            if (choiceIndex == nums.length) {
+                ArrayList<Integer> permutation = new ArrayList<Integer>();
+                for (int num : nums) {
+                    permutation.add(num);
+                }
```
ErdemT09, line 18: What are the approaches for converting `int[]` to `List<Integer>`?
Here are few other that I have found:
I don't think they can get any simpler than the current approach. It's good to know about them though.
`List<Integer> permutation = Arrays.stream(nums).boxed().collect(Collectors.toList());`
```
ArrayList<Integer> permutation = new ArrayList<Integer>();
Arrays.stream(nums).forEach(permutation::add);
```

src/main/java/algorithms/curated170/easy/Permutations.java
```diff
@@ -0,0 +1,34 @@
...
+                output.add(permutation);
+            }
+
+            for (int i = choiceIndex; i < nums.length; i++) {
+                swap(choiceIndex, i, nums);
+                backtrack(choiceIndex + 1, nums, output);
+                swap(i, choiceIndex, nums);
```
ErdemT09, line 25: We swap the number with itself too, which does not actually effect the result in a bad way.

ErdemT09 (approved):
The template is accurate as we can also observe it in other backtracking problems such as Generalized Abbreviation #43 and Factor Combinations #60.
Passing parameters vs using global variables is now something in the back of my head.
The code is pretty clean. Thank you for this contribution to algorithms.

altayhunoglu:
I have come across with a very similar idea in LeetCode too:
 https://leetcode.com/explore/learn/card/recursion-ii/472/backtracking/2793/

PR #306, opened 2021-06-08
487. Max Consecutive Ones II

### LINKED ISSUE #59: 487. Max Consecutive Ones II
https://leetcode.com/problems/max-consecutive-ones-ii/

Bir binary array (yani tüm değerleri 0 veya 1'lerden oluşan int[] nums arrayi) veriliyor.

Bu array'de "yalnızca bir adet" 0 değerini 1'e dönüştürme (flip) hakkı verilse, bu hakla oluşturabileceğiniz en uzun 1'ler dizisinin uzunluğu ne olurdu? 

Varsayılan kod:
```
class Solution {
    public int findMaxConsecutiveOnes(int[] nums) {
        
    }
}
```
image

ErdemT09:
Resolves #59 

### Points:
- Putting a 0 means that we are connecting 2 blocks of 1's.
- When we are inside a block of 1's, that means that we are either on the left or the united-right side of the block.
- When we hit a 0, that means that we move on the right block. The united block assumes the length left+1, we increment it with coming 1's if present after the 0. The left's length becomes 0 (It becomes the left of a new possible block).
- Each time, we check if the length of the united block is better than our previous best.

From these points, the algorithm follows.

For example, in the array `[1,1,0,1,1]`, we hit the 0. Here the united block assumes the length left+1 = 3. After encountering the other 1's, it gets the length 5.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/MaxConsecutiveOnes2.java
+++ b/src/main/java/algorithms/curated170/medium/MaxConsecutiveOnes2.java
@@ -1,7 +1,5 @@
 package algorithms.curated170.medium;
 
-import java.rmi.ConnectIOException;
-
 public class MaxConsecutiveOnes2 {
 
     public int findMaxConsecutiveOnes(int[] nums) {
@@ -20,7 +18,7 @@ public class MaxConsecutiveOnes2 {
                 curr = 0;
             }
 
-            best = Math.max(best, Math.max(curr, afterZeroCurr));
+            best = Math.max(best, afterZeroCurr);
         }
 
         return best;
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/MaxConsecutiveOnes2.java
```diff
@@ -0,0 +1,32 @@
+package algorithms.curated170.medium;
+
+public class MaxConsecutiveOnes2 {
+
+    public int findMaxConsecutiveOnes(int[] nums) {
```
altayhunoglu, line 5: Great:
image

altayhunoglu (approved):
Again, quite a clean explanation and self-expressive code.
Thanks, Erdem.

PR #308, opened 2021-06-08
Adds Backtracking solution: 47. Permutations II, moves Permutations to the right package.

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/307
 
 It is almost the same with: https://github.com/spiralgo/algorithms/pull/304
 
 This time, we add a "set" named `visited` to provide unique permutations, just because `nums` may include duplicate records this time.

### DIFF
--- a/src/main/java/algorithms/medium/Permutations2.java
+++ b/src/main/java/algorithms/medium/Permutations2.java
@@ -23,11 +23,12 @@ public class Permutations2 {
             output.add(permutation);
         }
 
-        Set<Integer> visited = new HashSet<>();
+        boolean[] visited = new boolean[21];
         for (int i = choiceIndex; i < nums.length; i++) {
-            if (!visited.add(nums[i])) {
+            if (visited[nums[i] + 10]) {
                 continue;
             }
+            visited[nums[i] + 10] = true;
             swap(choiceIndex, i, nums);
             backtrack(choiceIndex + 1, nums, output);
             swap(i, choiceIndex, nums);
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/medium/Permutations2.java
```diff
@@ -0,0 +1,44 @@
...
+    public void backtrack(int choiceIndex, int[] nums, List<List<Integer>> output) {
+        if (choiceIndex == nums.length) {
+            ArrayList<Integer> permutation = new ArrayList<Integer>();
+            for (int num : nums) {
+                permutation.add(num);
+            }
+            output.add(permutation);
+        }
+
+        Set<Integer> visited = new HashSet<>();
```
ErdemT09, line 26: We don't need a set. We know the range of numbers, which is -10 and 10. Therefore, instead, we can perhaps use an array of length 21.

src/main/java/algorithms/medium/Permutations2.java
```diff
@@ -0,0 +1,44 @@
...
+        for (int i = choiceIndex; i < nums.length; i++) {
+            if (!visited.add(nums[i])) {
+                continue;
+            }
```
ErdemT09, line 30:
```suggestion
        boolean[] visited = new boolean[21];
        for (int i = choiceIndex; i < nums.length; i++) {
            if (visited[nums[i] + 10]) {
                continue;
            }
            visited[nums[i] + 10] = true;
```

ErdemT09: Which is what we use in the palindrome permutation solution

altayhunoglu: Nice idea Erdem. Thanks. Suggestion commited.

ErdemT09 (approved):
Pretty legible code. This is indeed a nice algorithm for building up to other questions. Approved.

Issue #223, opened 2021-05-15
774. Minimize Max Distance to Gas Station


altayhunoglu:
Varsayılan kod:
```
class Solution {
    public double minmaxGasDist(int[] stations, int k) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380586-92d05180-b5eb-11eb-9ada-9c0b8c2cc9d1.png)
![image](https://user-images.githubusercontent.com/60903744/118380589-9b288c80-b5eb-11eb-917c-c478da1fa999.png)

ErdemT09:
https://codingcompetitions.withgoogle.com/kickstart/round/000000000019ffc7/00000000001d3f5b
Same question.

PR #311, opened 2021-06-10
774. Minimize Max Distance to Gas Station

ErdemT09:
Resolves: #223 

## Algorithm:
Using the given locations of the gas stations, we can deduce the path lengths between them.

### Approach 1-Discrete Priority Queue:
Let's say that we have two distances of 100 and 40. This equivalent to having 0 added stations between each. This also equivalent to dividing those distances by d+1=0+1=+1 . If we had put (d=1) a new station in the 100 length path, we would naturally put it into the middle to minimize the new penalty. This would reduce the path length to 100/2 = 50 = whole length / (d+1) =50
If we add a new station, we would naturally put into the path with the longest length. But now we have 2 paths with length 50 (from the divided 100) and a single path with length 40. Putting the new station inside one of the 50's wouldn't change anything as the penalty would remain 50. We can instead think of theses two paths of length 50 as parts of the original 100. So adding a new station means simply shifting the others, having the d of the 100 to be 2. 100/(d+1) ≈ 33.3. This equivalent to:
length of the divided parts * previous number of station (to restore it to the original value, e.g. 100) / number of new stations
This means that, each time, we choose the path with longest individual subpaths. After inserting the 2nd station to the 100-path, our next choice would be the path with length 40, as further dividing up the 100 into 25 wouldn't be profitable.
We can implement this using a priority queue.

### Approach 2-Improved, Continuous Priority Queue:
The paths between each given station can be treated as an interval.
The number of stations to be placed in that interval is `k` (the number of intervals that we can place) times the "proportion" of that interval, i.e. the amount of space that it occupies in terms of the total, from-start-to-end distance of the stations.
image
The most fair way of placing the stations in an interval (without the stations in-between) would have corresponded to placing them equidistantly. Normally, placing 4 stations in this interval of length 150 would have corresponded to placing them 37.5 units apart. 
Then, the amount of stations an interval has is either the ceil or the floor of its division with this value, in this case 37.5.
100 divided 37.5 approximately equals to 2.66. 
50 divided 37.5 approximately equals to 1.33. 

It's best to assume that the floor value is correct one at first, since we can't exactly predict it. This would lead to some new stations remaining unplaced.
Hence, the interval of 100 has 2 stations, giving them distance of 33.3. The interval of 50 has 1 station, giving them 25.

We put these intervals in a priority queue. We poll out the one with the largest distance and one-by-one place our remaining intervals. In this case, We have 1 interval left. We poll out the interval of 100 and add it one more station. Then we run out of new stations to end.

At the end, we poll out an interval from the priority queue. By the definition of our priority queue comparator and according to the problem statement that is the interval with the highest distance value.

### Approach 3-Binary Search:
We can see this question from another perspective:
Our return value is definitely something between 0 and the position of the last given station. We pick the middle value of these left and right values. We can try to calculate how many new station we need to put between the stations in order to get to this middle value. If the required number is larger than `k`, that means that the return value is larger than the middle value, else this means that we can add even more stations to get down to a smaller return value. If the first, the middle should become the minimum boundary, if the latter, the maximum boundary.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/MinimizeMaxDistanceToGasStation.java
+++ b/src/main/java/algorithms/curated170/hard/MinimizeMaxDistanceToGasStation.java
@@ -1,0 +1,37 @@
+package algorithms.curated170.hard;
+
+import java.util.Collections;
+import java.util.PriorityQueue;
+import java.util.TreeMap;
+
+public class MinimizeMaxDistanceToGasStation {
+
+    public double solve(int[] stats, int k)
+    {
+        double left = 0, right = stats[stats.length-1];
+        while(left + 1e-6 < right)
+        {
+            double mid = (left+right)/2;
+            int count = 0;
+            for(int i = 0; i<stats.length-1; i++)
+            {
+                count += (int)((stats[i+1]-stats[i])/mid);
+            }
+            if(count > k)
+            {
+                left = mid;
+            }
+            else
+            {
+                right = mid;
+            }
+        }
+        return right;
+    }
+
+    public static void main(String[] args) {
+        var solution = new MinimizeMaxDistanceToGasStation();
+
+        System.out.println(solution.solve(new int[]{1,2,3,4,5},4));
+    }
+}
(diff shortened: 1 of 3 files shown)


ErdemT09:
If have forgotten to change the method names. I will fix that when I can.

altayhunoglu (approved):
Thanks for your efforts and the brilliant explanation.

Visualization is always one step ahead and very useful, indeed.

PR #312, opened 2021-06-10
549. Binary Tree Longest Consecutive Sequence II

ErdemT09:
Resolves: #62 

## Algorithm:
Like many tree questions, this has a fairly simple recursive solution.

Like #300, we utilize the post-order DFS algorithm. First, we check the left node. If it gives us an ascending or descending sequence, we note down its length. Then, we check the right node. If it's ascending or descending and the left node also had the same relation, we choose the greatest. 
There can also be no sequence relationship between the node and either of its children, or the right node can overwrite the left's effect. The count of increase and decrease should then be declared initially as 0.
The longest sequence is then the maximum of the previous best the sum of the increasing arm, decreasing arm and 1 (the node itself).

Here is a visualization of this algorithm contributed by @altayhunoglu:
image

### DIFF
--- a/src/main/java/algorithms/curated170/medium/BinaryTreeLongestConsecutiveSequence2.java
+++ b/src/main/java/algorithms/curated170/medium/BinaryTreeLongestConsecutiveSequence2.java
@@ -1,0 +1,61 @@
+package algorithms.curated170.medium;
+
+import algorithms.datastructures.TreeNode;
+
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Thank you for this nice explanation and valuable contribution Erdem.
As we have collaborated from the start, it is already reviewed.
It is both the essence and byproduct of pair programming, I guess.

PR #313, opened 2021-06-11
361. Bomb Enemy

### LINKED ISSUE #65: 361. Bomb Enemy
https://leetcode.com/problems/bomb-enemy

Bir bomba ile öldürülebilecek maksimum düşman sayısı isteniyor.

Patlatılacak alan, iki boyutlu bir arrayde (char[][] grid) temsil ediliyor.

0 E 0 0 
E 0 W E 
0 E 0 0

Burada E: düşman (enemy), W: duvar (wall) , 0 ise boşluk anlamına geliyor.
Yukarıdaki gridde W'nin solundaki boş (0) yere bombayı yerleştirmek 3 düşmanı birden öldürüyor.

Varsayılan kod:

```
class Solution {
    public int maxKilledEnemies(char[][] grid) {
        
    }
}
```

image
image

ErdemT09:
Resolves #65 

## Algorithm:
Consider this grid for example, where brown tiles are the walls and the green ones are the enemies:
image
When we start the counting for each tile, we first start by counting the first column at the first row. Here, we count 2 for the row and 1 one the column. When we continue to count in the same row, we don't encounter any new walls, hence the count for this row should not change. We can keep the count of this row in a variable and change it only if we encounter a wall, otherwise we would be making unneeded checks.

The same things is valid for the columns. We have counted the column values up until their end or a wall as well. Since we are not directly iterating through the columns, we can keep the number of enemies that we have hit in an array for the columns. 
At the first column, for example, we can always reuse the information we have experienced from the first row, as we do not encounter any new walls their. 

In places like the 3rd row, our row counter will be reset when we come across a new wall.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/BombEnemy.java
+++ b/src/main/java/algorithms/curated170/medium/BombEnemy.java
@@ -40,7 +40,7 @@ public class BombEnemy {
     private boolean isStartOrAfterWall(int row, int col, boolean checkCol) {
         if (checkCol) {
             return col == 0 || grid[row][col - 1] == 'W';
-        } else {
+        } else checkRow: {
             return row == 0 || grid[row - 1][col] == 'W';
         }
     }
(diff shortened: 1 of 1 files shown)


ErdemT09:
The runtime is between 6 and 16 ms.
image

altayhunoglu (approved):
Very clean explanation and a self-expressive modular code Erdem. Thanks.

PR #314, opened 2021-06-13
314. Binary Tree Vertical Order Traversal

ErdemT09:
Resolves: #63 

## Algorithm:
In this problem, we have to return a lists of lists, where this list is sorted in terms of the positions of the nodes from left to right. The lists within this lists are what matters to the vertical order traversal.
So, one can think of the root being in the 0th column. The left child of root is then one unit to the left, the right child one unit to the right. This definition holds for all nodes in the tree. Every left node of a root is on a column 1 unit smaller and every right node is respectively on column at the 1 unit larger position.

### Approach 1-BFS & TreeMap
We implement the aformentioned method for the determining the "column values" of the nodes. If we traverse the tree layer by layer, we will add to each column list only in a vertical sorted order. The key of this TreeMap is then the column positions, while its values are just the vertically ordered lists.
For this, we also use an auxiliary class denoting the tree node and its column position.
At then end, returning a list of lists created from the TreeMap's values is sufficient.

### Approach 2-BFS & Range Preprocessing
In every tree, there is some maximum right position and some minimum left position. In all the locations therebetween, we are also guaranteed to find nodes since we are always going only 1 unit left or right.

While doing BFS, in order to avoid using a HashMap, we can shift our minimum value to an index 0 by initially denoting our root to have the column value `-1*minimum`.
Similar to the solution above, we place the root encapsulated in a class in a queue and place our node values in the list according to their column value. Row value is once again ignored because it is already taken care of by the level traversal of the BFS algorithm.
At the end, we directly return our list.

### Approach 3-DFS & Range Preprocessing & Sorting
If we apply DFS to a tree, we won't be able to freely add the node value to the corresponding list directly. So, instead of adding just the value at some column, we add a pair of row position and the node's value itself.

Here, similarly to the previous, we have also counted the minimum and maximum column values. We iterate over the range of these values, and retrieve the column pair list from the HashMap. We sort the column list according to the row values. We then create a new list add the node values from this sorted list to here. We then add this list to the output list. We eventually return the output list.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
+++ b/src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
@@ -17,7 +17,7 @@ public class BinaryTreeVerticalOrderTraversalBFS {
       return Collections.emptyList();
     }
     
-    nodeCount = 0;
+     
     int[] range = new int[] {0, 0};
     getRange(root, range, 0);
 
--- a/src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
+++ b/src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
@@ -54,8 +54,8 @@ public void getRange(TreeNode root, int[] range, int col) {
 }
 
   private class NodePosPair {
-    final TreeNode node;
-    final int x;
+    private final TreeNode node;
+    private final int x;
 
     void placeIntoMap(List<List<Integer>> columnTable) {
       columnTable.get(x).add(node.val);
--- a/src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
+++ b/src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
@@ -2,6 +2,7 @@ package algorithms.curated170.medium.braceexpansion;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import java.util.TreeMap;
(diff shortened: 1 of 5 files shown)


ErdemT09:
Appraoch 1:
image

Approach 2:
image

Approach 3:

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFSTreeNode.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.List;
+import java.util.TreeMap;
+
+import algorithms.datastructures.TreeNode;
+
+public class BinaryTreeVerticalOrderTraversalBFSTreeNode {
```
altayhunoglu, line 12: You mean **BinaryTreeVerticalOrderTraversalBFSTreeMap** I think.

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
```diff
@@ -0,0 +1,91 @@
...
+
+    nodeCount++;
+    range[0] = Math.min(range[0], col);
+    range[1] = Math.max(range[1], col);
+    
+    getRange(root.left, range, col - 1);
+    getRange(root.right, range, col + 1);
+}
+
+  private class NodePosPair {
```
altayhunoglu, line 56: We can define NodePosPair in an external class file to use it in multiple classes.

ErdemT09: In the two BFS solutions, the implementation is not the same. So, I think it's not worth bothering as we won't use  it again and it would create problems while testing.

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFSTreeNode.java
ErdemT09: Yes.

src/main/java/algorithms/datastructures/Pair.java
```diff
@@ -0,0 +1,15 @@
+package algorithms.datastructures;
```
ErdemT09, line 1: I have also added this. Might be useful at some point.

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
altayhunoglu: Yeah, of course. No problem.

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Collections;
+import java.util.Deque;
+import java.util.List;
+import java.util.TreeMap;
+
+import algorithms.datastructures.TreeNode;
+
+public class BinaryTreeVerticalOrderTraversalBFS {
+
+  int nodeCount;
```
altayhunoglu, line 14:
```suggestion

```

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
```diff
@@ -0,0 +1,91 @@
...
+  public List<List<Integer>> verticalOrder(TreeNode root) {
+    if (root == null) {
+      return Collections.emptyList();
+    }
+    
+    nodeCount = 0;
```
altayhunoglu, line 20:
```suggestion
     
```

src/main/java/algorithms/curated170/medium/binarytreeverticalordertraversal/BinaryTreeVerticalOrderTraversalBFS.java
```diff
@@ -0,0 +1,91 @@
...
+    
+    return columnsAtXPos;
+}
+
+public void getRange(TreeNode root, int[] range, int col) {
+    if (root == null) {
+        return;
+    }
+
+    nodeCount++;
```
altayhunoglu, line 48:
```suggestion
    nodeCount++;
```

altayhunoglu (approved):
Thank you, Erdem.
Nice explanation and clean code.

PR #315, opened 2021-06-13
419. Battleships in a Board

ErdemT09:
Resolves: https://leetcode.com/problems/battleships-in-a-board/

## Algorithm:
This is pretty simple:
Since the ships are only aligned vertically or horizontally, having a ship tile on the left tile or the tile above guarantees that we have counted that ship before, unless we are in the first column or row, then we always count the horizontal or the vertical, respectively.

### DIFF
--- a/src/main/java/algorithms/medium/BattleshipsInABoard.java
+++ b/src/main/java/algorithms/medium/BattleshipsInABoard.java
@@ -1,0 +1,30 @@
+package algorithms.medium;
+
+public class BattleshipsInABoard {
+
+    private final char BATTLESHIP = 'X';
+
+    public int countBattleships(char[][] board) {
+        int count = 0;
+        for (int i = 0; i < board.length; i++) {
+            for (int j = 0; j < board[0].length; j++) {
+
+                if (board[i][j] == BATTLESHIP && hasBattleshipBefore(i, j, false, board)
+                        && hasBattleshipBefore(i, j, true, board)) {
+                    count++;
+                }
+
+            }
+        }
+        return count;
+    }
+
+    private boolean hasBattleshipBefore(int i, int j, boolean checkLeft, char[][] board) {
+        if (checkLeft) {
+            return (j != 0 && board[i][j - 1] != 'X');
+        } else checkAbove: {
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/medium/BattleshipsInABoard.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private final char BATTLESHIP = 'X';
+
+    public int countBattleships(char[][] board) {
+        int count = 0;
+        for (int i = 0; i < board.length; i++) {
+            for (int j = 0; j < board[0].length; j++) {
+
+                if (board[i][j] == BATTLESHIP && !hasBattleshipBefore(i, j, false, board)
+                        && !hasBattleshipBefore(i, j, true, board)) {
```
altayhunoglu, line 13:
```suggestion
                if (board[i][j] == BATTLESHIP && !hasBattleshipBefore(i, j, board)
                        && !hasBattleshipBefore(i, j, board)) {
```

src/main/java/algorithms/medium/BattleshipsInABoard.java
```diff
@@ -0,0 +1,30 @@
...
+        return count;
+    }
+
+    private boolean hasBattleshipBefore(int i, int j, boolean checkLeft, char[][] board) {
+        if (checkLeft) {
+            return (j != 0 && board[i][j - 1] == BATTLESHIP);
+        } else checkAbove: {
+            return (i != 0 && board[i - 1][j] == BATTLESHIP);
+        }
+    }
```
altayhunoglu, line 28:
```suggestion
     private boolean hasBattleshipBefore(int i, int j, char[][] board) {
            return (j != 0 && board[i][j - 1] == BATTLESHIP) ||   (i != 0 && board[i - 1][j] == BATTLESHIP);
     }
}
```

ErdemT09: I wrote it out more explicitly, bu this is okay too.

altayhunoglu (approved):
Thank you for the valuable contribution, Erdem.

PR #316, opened 2021-06-13
562. Longest Line of Consecutive One in Matrix

ErdemT09:
Resolves: #68 

## Algorithm:
In each tile that is 1, we can count the lengths of the vertical, horizontal, diagonal and anti-diagonal lines at that point by incrementing the two position indices for the respective movement operation and counting until we hit a 0 or matrix bounds.

### Approach 1-Dynamic Programming:
Similar to #313, in a row, we can count the number of consecutive Ones, accumulating along the way. If we hit a 0, we reset this counter. 
We can also accumulate the horizontal consecutive length between the rows. Since we don't directly iterate through this path, we need an array to keep track of this.

For the diagonal, we need to check the previous column in the next row. When we get to the next row at there, the row has been incremented by 1 and the column is decremented by 1. Let's say that we are going to check this relationship at row = 1, and column 0. In the tile anti-diagonally before, the row was 0 and the column was 1. That means that in this relationship, the sum of row and column stay constant with the column always increasing. So, when storing our 1 at some position `i, j`, we store it at `i+j`.

For the diagonal, unlike in the horizontal, we accumulate for the next column, since that is going to be consecutive to the current one. Next diagonal of` i=2,  j=3` is `i=3,  j=4`. Here, `j-i` stands constant. Since, this can be smaller than 0, we add the size of the array to it to index it. We accumulate our count there.

We compare the current counts and pick the maximums among them and eventually return.

### Approach 2-Bit Flags:
Since we have a grid, we can modify it as well.
Let's say that the 2nd bit of some number at the grid denotes if it has been vertically visited. So, 3rd:horizontal, 4th:diagonal, 5th: anti-diagonal.
For example, this is some number from the grid in binary.
`010011`
It has been vertically and anti-diagonally visited. Since when we visit a tile, we also move on to its possible neighbor in the same direction, we don't explore through such notes in the directions where it has been bitwise-flagged.
When we are moving in some direction, we flag all the notes at the bit of that direction.
This way, we only use constant amount of space.

### Approach 3-Checking the Previous Tile:
Like in #315, having a previous tile being one in any of these four directions means that we have previously visited this current tile by the means of that previous tile, so we should not go in its direction. This approach also uses only constant space and doesn't modify the matrix.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/LongestLineOfConsecutiveOneInMatrix.java
+++ b/src/main/java/algorithms/curated170/medium/LongestLineOfConsecutiveOneInMatrix.java
@@ -1,0 +1,140 @@
+package algorithms.curated170.medium;
+
+import java.util.Arrays;
+
+public class LongestLineOfConsecutiveOneInMatrix {
+
+    final int VERTICAL = 0b10;
+    final int HORIZONTAL = 0b100;
+    final int DIAGONAL = 0b1000;
+    final int ANTID = 0b10000;
+
+    int[][] mat;
+    int n, m;
+    int longest = 0;
+
+    public int longestLine(int[][] mat) {
+        this.mat = mat;
+        n = mat.length;
+        m = mat[0].length;
+
+        for (int i = 0; i < n; i++) {
+            for (int j = 0; j < m; j++) {
+                control(i, j);
+            }
+        }
+        return longest;
+    }
+
+    void control(int i, int j) {
+        int curr = mat[i][j];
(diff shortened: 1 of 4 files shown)


ErdemT09:
Approach 1:
image

Approach 2:
7-8 ms
image

Approach 3: 
9-10 ms
image

altayhunoglu (approved):
Thank you, Erdem.
DP is the cleanest and relatable according to me.

PR #317, opened 2021-06-13
251. Flatten 2D Vector

ErdemT09:
Resolves: #67 

## Algorithm:
To check the value that we are going to return, we keep two pointer `i` and `j`. They are initially 0. At first, we return the first element of the first array. But what if the length of the array is only 1? Then we have to shift the row-pointer to the next array and set j back to 0. But some arrays can be empty. Then we shift the row-pointer until we found an array that is not empty. Otherwise, that means that we don't have any next elements.

With this logic, we can call the hasNext() method within the next() method. hasNext() takes care of both its normal defined function and shifting the row-pointer to an array that is not empty.


altayhunoglu (approved):
Nice explanation, lean code.
Thanks, Erdem.

PR #319, opened 2021-06-14
151. Reverse Words in a String

ErdemT09:
Resolves: https://leetcode.com/problems/reverse-words-in-a-string/

## Algorithm:
For this problem, the efficient way would be
1. Clear leading, trailing and other extra spaces
2. Reverse the entire string
3. Reverse each word within the string

We can do this in 2 ways. The first in using the built-in methods of our programming language. In this case, Java has a special method called `trim()` for deleting leading and trailing spaces from the string. We can then create an array of words by splitting the new string from its spaces. We can then reverse this array of words and join these words with a space between them. We just return the array of words converted into a single string.

The other method is doing this without the direct usage of the programming languages' utilities.
We first create a character array from our input string. We then remove the extra spaces having two indices and by overwriting the character at one if the next one is at a character that is not a space or is a space and does not have any spaces before it.

After that, we copy our character array to the previous character array at the length of the first, overwriter index. We reverse the string by having two pointers that converge into the middle.
After that, we reverse each word once again by having two pointers, with the second moving forward if we are on a letter. We then reverse this part of the string between these two indices. 
At the end, we just return this character array converted into a string.


ErdemT09:
Appraoch 1:
2-3 ms
image


Approach 2:
image

altayhunoglu (approved):
Thanks, Erdem.
Nothing to add for now.
We can add a Stack solution in the future.

PR #320, opened 2021-06-14
186. Reverse Words in a String II

ErdemT09:
Resolves: #70 

## Algorithm:
This is just a simpler version of #319. We just don't need to clean the extra spaces. So we can use the `reverse()` and `reverseEachWord()` methods from the previous solution. We can also use the built-in methods to separate the character array into an array of words, then reverse this, then convert this to a character array where we copy each element from this to the original.


ErdemT09:
Approach 1:
image


Appraoch 2, using built-in methods:
image

altayhunoglu (approved):
Thanks, Erdem. Quite clean codes.

PR #321, opened 2021-06-14
144. Binary Tree Preorder Traversal

ErdemT09:
Resolves: https://leetcode.com/problems/binary-tree-preorder-traversal/

## Algorithm:
This problem is pretty simple. We could either implement a DFS algorithm that is by definition a preorder traversal. 
As for the stack approach, we put the right node of a root first and then the left node, because the stack works on the principle "Last in, First out".

### DIFF
--- a/src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
+++ b/src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
@@ -1,4 +1,4 @@
-package algorithms.easy;
+package algorithms.easy.binarytreepreordertraversal;
 
 import java.util.ArrayList;
 import java.util.Collections;
--- a/src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
+++ b/src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
@@ -12,10 +12,6 @@ public class BinaryTreePreorderTraversal {
 
 	public List<Integer> preorderTraversal(TreeNode root) {
 
-		if (root == null) {
-			return Collections.emptyList();
-		}
-
 		DFSpreorder(root);
 		return preorder;
 	}
--- a/src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
+++ b/src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
@@ -21,16 +21,10 @@ public class BinaryTreePreorderTraversal {
 	}
 
 	private void DFSpreorder(TreeNode root) {
-
-		preorder.add(root.val);
-
-		if (root.left != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
```diff
@@ -0,0 +1,36 @@
...
+
+		preorder.add(root.val);
+
+		if (root.left != null) {
+			DFSpreorder(root.left);
+		}
+
+		if (root.right != null) {
+			DFSpreorder(root.right);
+		}
```
altayhunoglu, line 33:
```suggestion
          if (root == null) return;
          preorder.add(root.val);
          DFSpreorder(root.left);
          DFSpreorder(root.right);
```

src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversalStack.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Stack;
+
+import algorithms.datastructures.TreeNode;
+
+public class BinaryTreePreorderTraversalStack {
+
+	public List<Integer> preorderTraversal(TreeNode root) {
+		if (root == null) {
+			return Collections.emptyList();
+		}
```
altayhunoglu, line 15:
```suggestion
 
```

src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversal.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class BinaryTreePreorderTraversal {
+
+	List<Integer> preorder = new ArrayList<>();
+
+	public List<Integer> preorderTraversal(TreeNode root) {
+
+		if (root == null) {
+			return Collections.emptyList();
+		}
```
altayhunoglu, line 17:
```suggestion

```

src/main/java/algorithms/easy/binarytreepreordertraversal/BinaryTreePreorderTraversalStack.java
```diff
@@ -0,0 +1,35 @@
...
+
+		while (!stack.isEmpty()) {
+			TreeNode node = stack.pop();
+			preorder.add(node.val);
+			if (node.right != null) {
+				stack.push(node.right);
+			}
+			if (node.left != null) {
+				stack.push(node.left);
+			}
```
altayhunoglu, line 30:
```suggestion
            if(node == null) continue;
            preorder.add(node.val);
            stack.push(node.right);
	      stack.push(node.left);
```

ErdemT09: I think adding null values to the stack will be less efficient than one extra null-check.

altayhunoglu (approved):
Thanks, Erdem. Nice and clear code.

Issue #322, opened 2021-06-15
248. Strobogrammatic Number III


altayhunoglu:
Varsayılan kod:
```java
class Solution {
    public int strobogrammaticInRange(String low, String high) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/122051052-d38ad880-cdec-11eb-8b51-13ad4de1e7b8.png)

Issue #228, opened 2021-05-15
1088. Confusing Number II


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int confusingNumberII(int n) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380690-636e1480-b5ec-11eb-9920-939cee93cf08.png)
![image](https://user-images.githubusercontent.com/60903744/118380692-68cb5f00-b5ec-11eb-9cb7-cda55cebfd9e.png)
![image](https://user-images.githubusercontent.com/60903744/118380696-6cf77c80-b5ec-11eb-8c5d-3e5f260b859a.png)

PR #323, opened 2021-06-15
248. Strobogrammatic Number III

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/322

Even if the question is not in Curated 170, it is important as it will be a basis for: https://github.com/spiralgo/algorithms/issues/228

We use almost the same code with our solution in: https://github.com/spiralgo/algorithms/pull/305

This time: 
- We need to control if the number is in the `range` given with `low` and `high`.
- We will just count the numbers instead of adding them to a list.

### DIFF
--- a/src/main/java/algorithms/hard/StrobogrammaticNumber3.java
+++ b/src/main/java/algorithms/hard/StrobogrammaticNumber3.java
@@ -10,6 +10,10 @@ public class StrobogrammaticNumber3 {
         {'8', '8'}
     };
 
+    int highLen;
+    int lowLen;
+    int currLen;
+
     public int strobogrammaticInRange(String low, String high) {
 
         if (low == null || high == null || low.length() > high.length()
@@ -17,8 +21,12 @@ public class StrobogrammaticNumber3 {
             return 0;
         }
 
+        highLen = high.length();
+        lowLen = low.length();
+
         int count = 0;
-        for (int len = low.length(); len <= high.length(); len++) {
+        for (int len = lowLen; len <= highLen; len++) {
+            currLen = len;
             count += createNum(low, high, new char[len], 0, len - 1);
         }
         return count;
@@ -29,8 +37,7 @@ public class StrobogrammaticNumber3 {
         if (left > right) {
 
             String s = new String(numBuilder);
-            if ((numBuilder.length == low.length() && s.compareTo(low) < 0)
-                    || (numBuilder.length == high.length() && s.compareTo(high) > 0)) {
+            if ((currLen == lowLen && s.compareTo(low) < 0) || (currLen == highLen && s.compareTo(high) > 0)) {
                 return 0;
             } else {
                 return 1;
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/hard/StrobogrammaticNumber3.java
```diff
@@ -0,0 +1,57 @@
...
+
+        for (char[] pair : reflectives) {
+            char ch = pair[0];
+            numBuilder[left] = ch;
+            numBuilder[right] = pair[1];
+            if (numBuilder.length != 1 && numBuilder[0] == '0') {
+                continue;
+            } else if (left == right && (ch == '6' || ch == '9')) {
+                continue;
+            }
```
altayhunoglu, line 50:
```suggestion
   if ((numBuilder.length != 1 && numBuilder[0] == '0')  || (left == right && (ch == '6' || ch == '9'))) {
            continue;
           }
```

src/main/java/algorithms/hard/StrobogrammaticNumber3.java
```diff
@@ -0,0 +1,55 @@
...
+            numBuilder[right] = pair[1];
+   if ((numBuilder.length != 1 && numBuilder[0] == '0')  || (left == right && (ch == '6' || ch == '9'))) {
+            continue;
+           }
+            count += createNum(low, high, numBuilder, left + 1, right - 1);
+        }
+
+        return count;
+
+    }
```
ErdemT09, line 54: We can extract the length of the strings to fields to not call their methods again.

```suggestion
    int highLen;
    int lowLen;
    int currLen;

    public int strobogrammaticInRange(String low, String high) {

        if (low == null || high == null || low.length() > high.length()
                || (low.length() == high.length() && low.compareTo(high) > 0)) {
            return 0;
        }

        highLen = high.length();
        lowLen = low.length();

        int count = 0;
        for (int len = lowLen; len <= highLen; len++) {
            currLen = len;
            count += createNum(low, high, new char[len], 0, len - 1);
        }
        return count;

    }

    int createNum(String low, String high, char[] numBuilder, int left, int right) {
        if (left > right) {

            String s = new String(numBuilder);
            if ((currLen == lowLen && s.compareTo(low) < 0) || (currLen == highLen && s.compareTo(high) > 0)) {
                return 0;
            } else {
                return 1;
            }
        }

        int count = 0;

        for (char[] pair : reflectives) {
            char ch = pair[0];
            numBuilder[left] = ch;
            numBuilder[right] = pair[1];
            if ((currLen != 1 && numBuilder[0] == '0') || (left == right && (ch == '6' || ch == '9'))) {
                continue;
            }
            count += createNum(low, high, numBuilder, left + 1, right - 1);
        }

        return count;

    }
```

ErdemT09:
We can actually just convert the strings to character arrays and compare the arrays themselves, which would have been faster, but it's sometimes better to use what's built-in than trying to reinvent a better wheel.

altayhunoglu:
> Was einfacher für einen Mensch zu coden ist, ist manchmal besser als, was für eine Maschine schneller ist.
  
 You are right and thanks for the German translation.
 Mein guter kamerad. 
 It helps a lot.

ErdemT09 (approved):
Lean code following up from the prequel problem. Danke wieder für die Mitarbeit und Beitragleistung hier. Approved

PR #324, opened 2021-06-15
1088. Confusing Number II

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/228

As we understand here: https://github.com/spiralgo/algorithms/issues/156, if a number is generated using the numbers: 0, 1, 6, 8, 9 and :

 - if it does `not equal` itself when it is rotated 180 degrees, it is a `confusing number`. 
 
 - If it `equals` itself when it is rotated 180 degrees, it is a `strobogrammatic number`. 
 
 Therefore:

>  (Total numbers) - (strobogrammatic numbers) = (confusing numbers)

It means, we can use the code here https://github.com/spiralgo/algorithms/pull/323, and just subtract its output from the total count of all numbers generated using 0, 1, 6, 8, 9.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
+++ b/src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
@@ -21,11 +21,11 @@ public class ConfusingNumber2 {
     private int findTotal(String s) {
         if (s.length() == 0) return 1;
         char first = s.charAt(0);
-        int res = count(first) * (int) (Math.pow(5, s.length() - 1));
+        int reflectivePermutations = count(first) * (int) (Math.pow(5, s.length() - 1));
         if (first == '0' || first == '1' || first == '6' || first == '8' || first == '9') {
-            res += findTotal(s.substring(1));
+            reflectivePermutations += findTotal(s.substring(1));
         }
-        return res;
+        return reflectivePermutations;
     }
     private int count(Character c) {
         int res = 0;
--- a/src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
+++ b/src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
@@ -1,0 +1,42 @@
+package algorithms.curated170.hard;
+
+import algorithms.hard.StrobogrammaticNumber3;
+
+ 
+public class ConfusingNumber2 {
+   final char[][] reflectives = new char[][]{
+        {'0', '0'},
+        {'6', '9'},
+        {'1', '1'},
+        {'9', '6'},
+        {'8', '8'}
+    };
+
+    int highLen;
+    int lowLen;
+    int currLen;
+    
+    public int confusingNumberII(int n) {
+        String num = Integer.toString(n);
+        int res = findTotal(num);
+        StrobogrammaticNumber3 strobogrammaticNumber = new StrobogrammaticNumber3();
+        return res - strobogrammaticNumber.strobogrammaticInRange("1", num) - 1;
+    }
+    private int findTotal(String s) {
+        if (s.length() == 0) return 1;
+        char first = s.charAt(0);
+        int res = count(first) * (int) (Math.pow(5, s.length() - 1));
+        if (first == '0' || first == '1' || first == '6' || first == '8' || first == '9') {
+            res += findTotal(s.substring(1));
+        }
+        return res;
+    }
+    private int count(Character c) {
+        int res = 0;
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
```diff
@@ -0,0 +1,42 @@
...
+        {'0', '0'},
+        {'6', '9'},
+        {'1', '1'},
+        {'9', '6'},
+        {'8', '8'}
+    };
+
+    int highLen;
+    int lowLen;
+    int currLen;
```
altayhunoglu, line 17:
```suggestion

```

ErdemT09:
image
I don't understand how it's so efficient despite its use of strings and such.

src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
```diff
@@ -0,0 +1,38 @@
...
+    public int confusingNumberII(int n) {
+        String num = Integer.toString(n);
+        int res = findTotal(num);
+        StrobogrammaticNumber3 strobogrammaticNumber = new StrobogrammaticNumber3();
+        return res - strobogrammaticNumber.strobogrammaticInRange("1", num) - 1;
+    }
+    private int findTotal(String s) {
+        if (s.length() == 0) return 1;
+        char first = s.charAt(0);
+        int res = count(first) * (int) (Math.pow(5, s.length() - 1));
```
ErdemT09, line 24: What is being done here?

altayhunoglu: findTotal is to calculate the total count of the numbers which can be generated using those five numbers: 0, 1, 6, 8, 9.
It is why we use 5 to calculate the permutation.

As the "s" is a limit for us, we have to calculate numbers less than "s".
In order to do that: count(first) returns the numbers less than the parameter (first) from this set: 0, 1, 6, 8, 9.

findTotal is recursive as we should do the calculation for each digit.

ErdemT09: When considering a number like 9999, we ignore the permutations that start with 9. But, when we calculate the next digits, we are actually calculating the number of permutations that start with 9. Pretty clear now.

src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
```diff
@@ -0,0 +1,38 @@
...
+        if (first == '0' || first == '1' || first == '6' || first == '8' || first == '9') {
+            res += findTotal(s.substring(1));
+        }
+        return res;
+    }
+    private int count(Character c) {
+        int res = 0;
+        for (char[] p : reflectives) {
+            if (p[0] < c) res += 1;
```
ErdemT09, line 33:
```suggestion
            if (p[0] < c) res ++;
```

src/main/java/algorithms/curated170/hard/ConfusingNumber2.java
```diff
@@ -0,0 +1,38 @@
...
+        return res - strobogrammaticNumber.strobogrammaticInRange("1", num) - 1;
+    }
+    private int findTotal(String s) {
+        if (s.length() == 0) return 1;
+        char first = s.charAt(0);
+        int res = count(first) * (int) (Math.pow(5, s.length() - 1));
+        if (first == '0' || first == '1' || first == '6' || first == '8' || first == '9') {
+            res += findTotal(s.substring(1));
+        }
+        return res;
```
ErdemT09, line 28: Writing it a bit more explicity.
```suggestion
        int reflectivePermutations = count(first) * (int) (Math.pow(5, s.length() - 1));
        if (first == '0' || first == '1' || first == '6' || first == '8' || first == '9') {
            reflectivePermutations += findTotal(s.substring(1));
        }
        return reflectivePermutations;
```

ErdemT09 (approved):
It's a good method of problem solving to reduce a problem to another one. I like how that's being done here. 
Approved.

altayhunoglu:
> It's a good method of problem solving to reduce a problem to another one. I like how that's being done here.
> Approved.

Vielen Dank für die Codeüberprüfung und für deine wertvolle Zusammenarbeit.

Ich rufe bewusst eine Methode in einer anderen Klasse auf, um ein Beispiel für unsere Idee des "Lego-Bausteine" zu geben.

PR #325, opened 2021-06-16
277. Find the Celebrity

ErdemT09:
Resolves: #75 
## Algorithm:
We begin with the assumption that 0 is the celebrity.
If 0 does not know 1, this does not contradict this fact.
If 0 knows 1, this means that 0 can't be the celebrity. We can assume that 1 is the celebrity until disproven.
⇒ If 1 does not know 2, this does not contradict our current assumption.
If 1 does know 2, 1 can't be the celebrity, we can now assume that 2 is the celebrity.
Hence, at the end, the person, whom we assume is the celebrity, is the person that does not know anybody after the index.

What we have not proven is that the celebrity does not know anybody before. If the person knows anybody before, this means that there is no celebrity and we return -1;

If the celebrity in fact does not know anybody else, what we also need to prove is that everyone knows the celebrity. If we encounter someone that does not know the person, we return -1. If we successfully go through asking everyone and getting true values, we return this celebrity.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/FindTheCelebrity.java
+++ b/src/main/java/algorithms/curated170/medium/FindTheCelebrity.java
@@ -1,0 +1,54 @@
+package algorithms.curated170.medium;
+
+public class FindTheCelebrity {
+
+    public int findCelebrity(int n) {
+        int celeb = 0;
+        for (int p = 0; p < n; ++p)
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Die schlanken Lösungen gefallen mir sehr gut.
Danke.

PR #327, opened 2021-06-17
742. Closest Leaf in a Binary Tree

ErdemT09:
Resolves: #73 

## Algorithm:
What is meant by "closest" here is not the value but the distance, so actually the "nearest". 

There are 2 ways of approaching this problem:

### 1-Create an Undirected Graph from the Tree:
If the binary tree was an undirected graph, the nearest leaf simply would have been the first leaf we reach when start a BFS from the target node.
For this, we have to create a graph from the tree using a map data structure. In a DFS algorithm for this, we also pass down the parent node so that we could connect that to the graph at the node, to create an undirected graph.
After that, we create a queue where we add our target node and expand onto the graph from here. The first leaf will be the nearest leaf, so we return it.

### 2-Find the Node and Search for the Leaves:

When we find the target node, the closest leaf it is first assumed to be one of it's own leaves. Let's say this target is the left node of its parent. Then we might check the right branch of the node to find a nearer leaf. If the parent is the right node of its parent, then we might check the left branch of this parent's parent to find a nearer leaf. 
Basically, the algorithm is 

- check left/right
- when found check the node's own leaves
- if found the node at left/right, check the leaves on the right/left

### DIFF
--- a/src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeGraphBFS.java
+++ b/src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeGraphBFS.java
@@ -11,72 +11,55 @@ import algorithms.datastructures.TreeNode;
 
 public class ClosestLeafInABinaryTreeGraphBFS {
 
-  Map<TreeNode, List<TreeNode>> graph;
-  int k;
-  TreeNode nodeK = null;
+  Map<Integer, List<TreeNode>> graph = new HashMap<>();
 
   public int findClosestLeaf(TreeNode root, int k) {
     if (root.left == null && root.right == null) {
       return k;
     }
 
-    this.k = k;
-    graph = new HashMap<>();
     createGraphWithDFS(root, null);
 
-    return shortestPathBFS(root);
+    return shortestPathBFS(root, k);
   }
 
-  private int shortestPathBFS(TreeNode root) {
-    Deque<TreeNode> queue = new ArrayDeque<>();
+  private int shortestPathBFS(TreeNode root, int k) {
+    Deque<Integer> queue = new ArrayDeque<>();
     boolean[] seen = new boolean[1001];
 
-    queue.offer(nodeK);
+    queue.offer(k);
     seen[k] = true;
 
     while (!queue.isEmpty()) {
-      TreeNode node = queue.poll();
-      if (node != null) {
-        if (graph.get(node).size() <= 1 && node != root) {
-          return node.val;
-        }
-        for (TreeNode adjacentNode : graph.get(node)) {
-          if (!seen[adjacentNode.val]) {
-            seen[adjacentNode.val] = true;
-            queue.add(adjacentNode);
-          }
+      Integer node = queue.poll();
+
+      List<TreeNode> children = graph.get(node);
+      if (node != root.val && children.size() <= 1) {
+        return node;
+      }
+      for (TreeNode adjacentNode : children) {
+        if (!seen[adjacentNode.val]) {
+          seen[adjacentNode.val] = true;
+          queue.add(adjacentNode.val);
         }
       }
+
     }
 
-    throw null;
+    return -1;
   }
 
   public void createGraphWithDFS(TreeNode node, TreeNode parent) {
     if (node != null) {
-      if (node.val == k) {
-        nodeK = node;
-      }
-      graph.put(node, new ArrayList<>());
+
+      graph.put(node.val, new ArrayList<>());
       if (parent != null) {
-        graph.get(node).add(parent);
-        graph.get(parent).add(node);
+        graph.get(node.val).add(parent);
+        graph.get(parent.val).add(node);
       }
 
       createGraphWithDFS(node.left, node);
       createGraphWithDFS(node.right, node);
     }
-  }
-
-  public static void main(String[] args) {
-    var solution = new ClosestLeafInABinaryTreeGraphBFS();
-
-    TreeNode tn1 = new TreeNode(1);
-    TreeNode t0 = new TreeNode(2, null, tn1);
-    TreeNode t1 = new TreeNode(5);
-    TreeNode t2 = new TreeNode(4, null, t1);
-    TreeNode t3 = new TreeNode(3, t0, t2);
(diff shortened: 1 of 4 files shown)


src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeGraphBFS.java
```diff
@@ -0,0 +1,82 @@
...
+    while (!queue.isEmpty()) {
+      TreeNode node = queue.poll();
+      if (node != null) {
+        if (graph.get(node).size() <= 1 && node != root) {
+          return node.val;
+        }
+        for (TreeNode nei : graph.get(node)) {
+          if (!seen[nei.val]) {
+            seen[nei.val] = true;
+            queue.add(nei);
```
ErdemT09, line 46:
```suggestion
        for (TreeNode adjacentNode : graph.get(node)) {
          if (!seen[adjacentNode.val]) {
            seen[adjacentNode.val] = true;
            queue.add(adjacentNode);
```

ErdemT09:
Approach 1:
image

Approach 2:
image

altayhunoglu:
For the future audience, I can add this simple but deceptive idea:

Asking "`Why should we create an undirected graph instead of just directly traversing with BFS (in level-order) over the tree?`" may seem favorable at the first glance.

 If the target is also the root, this question would have been valid.
 But it is a rare case. What if it is not the root, like in this question:
 
image

As the target "2" is not the root, we cannot use the tree directly to BFS-traverse as we can't go back to the "parent" node.
It is why we create an undirected graph to destroy that kind of parent-child hierarchy, in some sense.

src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeSearchOtherBranch.java
```diff
@@ -0,0 +1,61 @@
...
+    findTarget(root, k);
+    return val;
+  }
+
+  private int findTarget(TreeNode root, int target) {
+    if (root == null) {
+      return -1;
+    }
+    if (root.val == target) {
+      seacrhClosestLeaf(root, 0);
```
altayhunoglu, line 20:
```suggestion
      searchClosestLeaf(root, 0);
```

src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeSearchOtherBranch.java
```diff
@@ -0,0 +1,61 @@
...
+      right = findTarget(root.right, target);
+      if (right > 0) {
+        seacrhClosestLeaf(root.left, ++right);
+      }
+    }
+
+    return Math.max(left, right);
+  }
+
+  private void seacrhClosestLeaf(TreeNode root, int dist) {
```
altayhunoglu, line 37:
```suggestion
  private void searchClosestLeaf(TreeNode root, int dist) {
```

src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeGraphBFS.java
```diff
@@ -0,0 +1,82 @@
...
+
+    TreeNode tn1 = new TreeNode(1);
+    TreeNode t0 = new TreeNode(2, null, tn1);
+    TreeNode t1 = new TreeNode(5);
+    TreeNode t2 = new TreeNode(4, null, t1);
+    TreeNode t3 = new TreeNode(3, t0, t2);
+    TreeNode root = new TreeNode(9, null, t3);
+
+  }
+}
```
altayhunoglu, line 82: The code has no problem I think.
Nevertheless, we can think about using `Integer` keys for the graph.
It takes 3ms in this way.
(We can also use `List<TreeNode>[] graph`, but the map is fine I think.)

```java
class Solution {

  Map<Integer, List<TreeNode>> graph = new HashMap<>();
 

  public int findClosestLeaf(TreeNode root, int k) {
    if (root.left == null && root.right == null) {
      return k;
    }
 
    createGraphWithDFS(root, null);

    return shortestPathBFS(root, k);
  }

  private int shortestPathBFS(TreeNode root, int k) {
    Deque<Integer> queue = new ArrayDeque<>();
    boolean[] seen = new boolean[1001];

    queue.offer(k);
    seen[k] = true;

    while (!queue.isEmpty()) {
      Integer node = queue.poll();
      
        List<TreeNode> children = graph.get(node);  
        if (node != root.val && children.size() <= 1) {
          return node;
        }
        for (TreeNode adjacentNode : children) {
          if ( !seen[adjacentNode.val]) {
            seen[adjacentNode.val] = true;
            queue.add(adjacentNode.val);
          }
        }
      
    }

   return -1;
  }

  public void createGraphWithDFS(TreeNode node, TreeNode parent) {
    if (node != null) {
    
      graph.put(node.val, new ArrayList<>());
      if (parent != null) {
        graph.get(node.val).add(parent);
        graph.get(parent.val).add(node);
      }

      createGraphWithDFS(node.left, node);
      createGraphWithDFS(node.right, node);
    }
  }
}
```

src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeSearchOtherBranch.java
ErdemT09: Fixed.

src/main/java/algorithms/curated170/medium/closestleafinabinarytree/ClosestLeafInABinaryTreeGraphBFS.java
ErdemT09: We can do such thins here because the Node values are unique. That was also the case with the boolean array `seen`.

ErdemT09: Changes added.

altayhunoglu (approved):
Good quality and readable code.
Thank you for your valuable effort, Erdem.

PR #328, opened 2021-06-17
291. Word Pattern II

ErdemT09:
Resolves: #74 

## Algorithm:
Here are few points for choosing substrings to match to a pattern through a recursive backtracking algorithm:
Let's say that we are the start of some string str and we want to try out some substring to match to our pattern of length n. The length of the remaining, other substring in the string has to be greater or equal to n-1, since there will be that many characters left out to match to some substring in the str, at least of the length 1. Every time we try a substring, we move on to the next char in the pattern too. If this possible match does not work, we shorten our match. If no substrings in the current bounds works, this means that it is impossible to match the string. If this happens at the start, it means that there are no possibilities at all. If this happens at some backtracking branch, this means that the branch is false.

If we traverse either the entire str or the pattern and have not completed both of them at the same, this means that the current possibility of matches is false, one of the base cases. If we have completed the both, that means that we have found a true possibility, returning true all the way up to the start backtracking.

If we are at some index in the pattern and the string that we have assigned to this char is longer than the rest of the string, or the string does not have this match in its current substring, or these are true but assuming this for the rest of the string ends up with false because this or the previous base case is false, this means that it is false.

We need to use a set of strings too, so we don't map the same substring to another character in the pattern. We can map characters to strings using a simple array of strings though. These guarantee a "bijective mapping".

The algorithm builds itself upon these ideas.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/WordPattern2.java
+++ b/src/main/java/algorithms/curated170/medium/WordPattern2.java
@@ -1,0 +1,100 @@
+package algorithms.curated170.medium;
+
+import java.util.HashSet;
+
+public class WordPattern2 {
+
+    String[] mapMatches;
+    HashSet<String> setMatches;
+    String pattern, str;
+
+    public boolean wordPatternMatch(String pattern, String str) {
+        mapMatches = new String[26]; // mapping of characters 'a' - 'z'
+        setMatches = new HashSet<>(); // mapped result of 'a' - 'z'
+        this.pattern = pattern;
+        this.str = str;
+        return backtrackPatternMatch(0, str.length() - 1, 0, pattern.length() - 1);
+    }
+
+    private boolean backtrackPatternMatch(int start, int end, int pStart, int pEnd) {
+
+        if (pStart == pEnd + 1 && start == end + 1) {
+            return true;
+        } else if ((pStart > pEnd) || (start > end)) {
+            return false;
+        }
+
+        char ch = pattern.charAt(pStart);
+        int chIdx = letterIndex(ch);
+        String matched = mapMatches[chIdx];
+
+        if (matched != null) {
+            return isValidMatch(start, end, pStart, pEnd, matched);
+        }
+
+        int unmatchedEndPos = end;
+
+        for (int i = pEnd; i > pStart; i--) {
+            unmatchedEndPos -= patternsToMatch(i);
+        }
+
+        for (int i = start; i <= unmatchedEndPos; i++) {
+
+            matched = str.substring(start, i + 1);
+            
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/WordPattern2.java
```diff
@@ -0,0 +1,111 @@
...
+    private int patternsToMatch(int i) {
+        return mapMatches[letterIndex(pattern.charAt(i))] == null ? 1
+                : mapMatches[letterIndex(pattern.charAt(i))].length();
+    }
+
+    private boolean isValidMatch(int start, int end, int pStart, int pEnd, String matched) {
+        int lengthMatchedStr = matched.length();
+        return start + lengthMatchedStr <= end + 1 && matched.equals(str.substring(start, start + lengthMatchedStr))
+                && backtrackPatternMatch(start + lengthMatchedStr, end, pStart + 1, pEnd);
+    }
```
altayhunoglu, line 81:
```suggestion
    private boolean isValidMatch(int start, int end, int pStart, int pEnd, String matched) {
        return str.startsWith(matched, start) && backtrackPatternMatch(start + matched.length(), end, pStart + 1, pEnd);
    }
```

src/main/java/algorithms/curated170/medium/WordPattern2.java
```diff
@@ -0,0 +1,111 @@
...
+        for (int i = pEnd; i > pStart; i--) {
+            unmatchedEndPos -= patternsToMatch(i);
+        }
+        return unmatchedEndPos;
+    }
+
+    private int patternsToMatch(int i) {
+        return mapMatches[letterIndex(pattern.charAt(i))] == null ? 1
+                : mapMatches[letterIndex(pattern.charAt(i))].length();
+    }
```
altayhunoglu, line 75:
```suggestion

```

I am not sure if it is any better than:
    
`str.length() - (pattern.length() - pStart)`

ErdemT09: If we have "abc"  at the end of the string, each is perhaps assigned to patterns length 10. We would then have to subtract  a total of 30 length from the end position.

altayhunoglu: There is this constraint: `1 <= pattern.length, s.length <= 20`

ErdemT09: Some patterns are already matched to substrings within the string. So assuming that these already matched patterns are of length 1 is not right.

altayhunoglu (approved):

PR #330, opened 2021-06-17
1059. All Paths from Source Lead to Destination

ErdemT09:
Resolves: #77 

## Algorithm:
It's good to know about the nomenclature of directed graphs here: https://en.wikipedia.org/wiki/Directed_graph
For solving this problem in anyway, we need to turn this given arcs into a directed graph.  For this, we first iterate over the arcs and count the outdegrees of the respective predecessor vertices. With this, we create a 2 dimensional array where the number of arrays is the given number `n` and the length of the each array is their respective outdegrees, otherwise we would exceed the memory (see the solution comment below). We then fill up each array with their successor vertices.

While creating the graph, if the outdegree of the destination node is not 0, this means that there is a path that is not leading to this target, it continues after the target. We should then check and return 0. If the source node has 0 arcs coming out of it, this means that the number of paths that lead to destination is 0. This means that we can't reach the target unless the source is the destination. If the source is the target, that would mean in this case that all paths from the source (0 paths) always end up in the target.

We than start a DFS from the source node. Here, we mark a node visited and then travel to its successor nodes. If we come across some visited node, this means that we have a cycle and should return false. If we end up in a node that has an outdegree of 0 but isn't destination, we return false. After traveling to and through all the successors of a node, we mark it unvisited and return true if we weren't interrupted by the aforementioned conditions.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/AllPathsfromSourceLeadToDestination.java
+++ b/src/main/java/algorithms/curated170/medium/AllPathsfromSourceLeadToDestination.java
@@ -9,14 +9,14 @@ public class AllPathsfromSourceLeadToDestination {
     boolean[] visiting;
     int dest;
 
-    public boolean leadsToDestination(int n, int[][] edges, int source, int destination) {
+    public boolean leadsToDestination(int n, int[][] arcs, int source, int destination) {
         this.dest = destination;
         visiting = new boolean[n];
 
-        byte graphCreation = makeGraph(edges, n, source);
-        if (graphCreation == DEST_HAS_PATH) {
+        byte graphCreation = makeGraph(arcs, n, source);
+        if (graphCreation == DEST_HAS_ARC) {
             return false;
-        } else if (graphCreation == SOURCE_HAS_NO_PATH) {
+        } else if (graphCreation == SOURCE_HAS_NO_ARC) {
             return source == destination;
         }
 
@@ -24,20 +24,20 @@ public class AllPathsfromSourceLeadToDestination {
     }
 
     final byte GRAPH_CREATED = 1;
-    final byte DEST_HAS_PATH = 0;
-    final byte SOURCE_HAS_NO_PATH = 2;
+    final byte DEST_HAS_ARC = 0;
+    final byte SOURCE_HAS_NO_ARC = 2;
     
-    byte makeGraph(int[][] edges, int n, int source) {
+    byte makeGraph(int[][] arcs, int n, int source) {
         int[] outDegree = new int[n];
-        for (int[] edge : edges) {
-            outDegree[edge[0]]++;
+        for (int[] arc : arcs) {
+            outDegree[arc[0]]++;
         }
 
         if (outDegree[dest] > 0) {
-            return DEST_HAS_PATH;
+            return DEST_HAS_ARC;
         }
         if (outDegree[source] == 0) {
-            return SOURCE_HAS_NO_PATH;
+            return SOURCE_HAS_NO_ARC;
         }
 
         graph = new int[n][];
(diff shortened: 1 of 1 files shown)


ErdemT09:
There is also this solution that entirely skips counting the outdegree of nodes and creating the graph accordingly. It passes 21 tests but exceeds the memory limit at n=10000.

```java
class Solution {
    
      int[][] graph;
    boolean[] visiting;
    int dest;
    int n;

    public boolean leadsToDestination(int n, int[][] edges, int source, int destination) {
        this.n = n;
        this.dest = destination+1;
        visiting = new boolean[n+1];

        byte graphCreation = makeGraph(edges, n, source);
        if (graphCreation == DEST_HAS_PATH) {
            return false;
        } else if (graphCreation == SOURCE_HAS_NO_PATH) {
            return source == destination;
        }

        return hasAllPathsToDestination(source+1);
    }

    final byte GRAPH_CREATED = 1;
    final byte DEST_HAS_PATH = 0;
    final byte SOURCE_HAS_NO_PATH = 2;

    byte makeGraph(int[][] edges, int n, int source) {
        
        graph = new int[n+1][n+1];
        int[] nextDirectSuccessorIndex = new int[n+1];
        for (int[] edge : edges) {
            graph[edge[0]+1][nextDirectSuccessorIndex[edge[0]+1]++] = edge[1]+1;
        }
        if(graph[source+1][0] == 0)
        {
            return SOURCE_HAS_NO_PATH;
        }
        if(graph[dest][0] != 0)
        {
            return DEST_HAS_PATH;
        }
        return GRAPH_CREATED;
    }

    boolean hasAllPathsToDestination(int currVertex) {

        if (graph[currVertex][0] == 0 && currVertex != dest) {
            return false;
        }
        visiting[currVertex] = true;

        for (int i = 0; i<n+1; i++) {
            
            int directedSuccessor = graph[currVertex][i];
            if(directedSuccessor == 0)
            {
                break;
            }
            if (visiting[directedSuccessor] || !hasAllPathsToDestination(directedSuccessor)) {
                return false;
            }
        }

        visiting[currVertex] = false;
        return true;
    }
}
```

ErdemT09:
I hope this images illustrates how some cases have the algorithm evaluate `false`:
image

altayhunoglu:
> I hope this images illustrates how some cases have the algorithm evaluate `false`:
> 
> image
> 
> 

Thanks for the image Erdem.
As I learn a lot from your PR's I will review all the PR's carefully. It is why it may take some time. Sorry about that.

ErdemT09:
> Thanks for the image Erdem.
> As I learn a lot from your PR's I will review all the PR's carefully. It is why it may take some time. Sorry about that.

If there are any ways of improving my explanation techniques, please let me know.

altayhunoglu:
> > Thanks for the image Erdem.
> > As I learn a lot from your PR's I will review all the PR's carefully. It is why it may take some time. Sorry about that.
> 
> If there are any ways of improving my explanation techniques, please let me know.

 Sure. We can talk about this kind of issue whenever you want.
 Just remind me whenever you would like to talk.

src/main/java/algorithms/curated170/medium/AllPathsfromSourceLeadToDestination.java
```diff
@@ -0,0 +1,72 @@
...
+        } else if (graphCreation == SOURCE_HAS_NO_ARC) {
+            return source == destination;
+        }
+
+        return hasAllPathsToDestination(source);
+    }
+
+    final byte GRAPH_CREATED = 1;
+    final byte DEST_HAS_ARC = 0;
+    final byte SOURCE_HAS_NO_ARC = 2;
```
altayhunoglu, line 28:
```suggestion
 
```

src/main/java/algorithms/curated170/medium/AllPathsfromSourceLeadToDestination.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class AllPathsfromSourceLeadToDestination {
+
+    int[][] digraph;
+    boolean[] visiting;
+    int dest;
+
```
altayhunoglu, line 11:
```suggestion
    final static byte GRAPH_CREATED = 1;
    final static byte DEST_HAS_ARC = 0;
    final static byte SOURCE_HAS_NO_ARC = 2;
    
```

altayhunoglu (approved):
Very readable and clean code.
Nice usage of constant variables as a signal flag to determine what to do.

Thanks Erdem.

PR #331, opened 2021-06-18
624. Maximum Distance in Arrays

ErdemT09:
Resolves: #80 

## Algorithm:

Since each individual array (LeetCode updated them to lists, but the same is valid for any such data structure) is sorted in ascending order, the maximum distance will be between the first element of some array and the last element of some other array. For this, we simply keep the minimum and maximum values in the arrays that we previously traversed. We compare the last and the first elements of the array respectively with the maximum and minimum values and update the maximum distance if possible. If the first or the last element is respectively smaller than the previous minimum or greater than the previous maximum, we update these values as well and continue to the other list.


altayhunoglu (approved):
Nice explanation and clean code.
Thanks, Erdem.

PR #332, opened 2021-06-18
261. Graph Valid Tree

### LINKED ISSUE #76: 261. Graph Valid Tree
https://leetcode.com/problems/graph-valid-tree

 validTree(int n, int[][] edges) fonksiyonu içerisinde n= node sayısı, edges ise bu node'ların birbirleri ile olan bağlantılarını içeriyor.

Verilen n ve edges'in, geçerli bir ağaç oluşturup oluşturmadığı soruluyor.

Varsayılan kod:
```
class Solution {
    public boolean validTree(int n, int[][] edges) {
        
    }
}
```

image
image
image

### LINKED ISSUE #261: 1198. Find Smallest Common Element in All Rows
Varsayılan kod:
```
class Solution {
    public int smallestCommonElement(int[][] mat) {
        
    }
}

```
image
image

ErdemT09:
Resolves #76.

## Algorithm:
There are few conditions of a such a given graph being a tree. It must
- be acyclic,
- have n-1 edges for n nodes,
- be connected,
- satisfy the condition that no connected node should have the same root.

The second condition is always easy to check at the start by comparing the number of edges and the given number `n`. If it passes the test, this means that some edges of some unconnected components lead cycles. For the rest, we have two strategies:

### Approach 1-DFS:
First, we convert the edges into a graph. This graph is just the undirected version from the same type of graph creation as in #330. After that, we visit some first node with the index 0 and mark this visited. We then visit each of its children. While traversing as such, if we ever come across a child that is already visited, that means we have a cycle and have to return false. We also count the number of nodes that we have traversed and compare to n if the DFS is successful in order to assure that the Tree is connected and the tree that we have looked at is not just some acyclic part of the disconnected graph components. 

### Approach 2-Union-Find:
This one is similar to previous Union-Find questions, but I will compare it to #329 because of its relevancy. If an edge connects two nodes that have the same root, this means that such an edge can't be the part of a valid tree. For checking this, we can simply implement a basic Union-Find simple additionally with path compression and ranking. If we can do union all nodes of all the edges, this means that we have a valid tree and can return true. Finding the same root also means finding a cycle, which also proves that the components are disconnected.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/graphvalidtree/GraphValidTreeDFS.java
+++ b/src/main/java/algorithms/curated170/medium/graphvalidtree/GraphValidTreeDFS.java
@@ -1,0 +1,49 @@
+package algorithms.curated170.medium;
+
+public class GraphValidTreeDFS {
+
+    int[][] graph;
+    boolean[] visited;
+    int nodeCount = 0;
+
+    public boolean validTree(int n, int[][] edges) {
+        if (edges.length != n - 1) {
+            return false;
+        }
+
+        visited = new boolean[n];
+        createGraph(n, edges);
+        return isValidTreeNode(0, -1) && nodeCount == n;
+    }
+
+    private boolean createGraph(int n, int[][] edges) {
+        int[] adjacentCount = new int[n];
+        for (int[] edge : edges) {
+            adjacentCount[edge[0]]++;
+            adjacentCount[edge[1]]++;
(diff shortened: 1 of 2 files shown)


altayhunoglu (approved):
Nice explanation and a good way to yielding from experiences we had in the past.
Thanks, Erdem.

Issue #213, opened 2021-05-15
1183. Maximum Number of Ones


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int maximumNumberOfOnes(int width, int height, int sideLength, int maxOnes) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380270-1dfc1800-b5e9-11eb-976f-5208367172f3.png)
![image](https://user-images.githubusercontent.com/60903744/118380278-26545300-b5e9-11eb-8d36-f8a86185d30d.png)

PR #336, opened 2021-06-20
1183. Maximum Number of Ones

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/213

- We can imagine a **square** at the top left of the matrix.
- We can assume that the matrix itself is a product of copying this square multiple times and cutting the edges according to the matrix size.
- Then we can assign an id for each tile of the square. (Just like we have done in https://github.com/spiralgo/algorithms/pull/329 with **getNodeId()**)

We need to count how many times each `tile id` could repeat (if the tile would have been selected to set 1 in it)
Then we select `top n` (=maxOnes) tiles that repeat most in order to get the `maximum benefit`.
(We could also use a PQ instead of sorting the array.)

IMG-1670

### DIFF
--- a/src/main/java/algorithms/curated170/hard/MaximumNumberofOnes.java
+++ b/src/main/java/algorithms/curated170/hard/MaximumNumberofOnes.java
@@ -1,0 +1,38 @@
+package algorithms.curated170.hard;
+
+import java.util.Arrays;
+
+ 
+public class MaximumNumberofOnes {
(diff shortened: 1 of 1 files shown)


ErdemT09 (approved):
This seems to me like a combination of graph and array problems. The image and the concise explanation helps a lot with the understanding. Once again, thank you for your contribution. 
Approved.

Issue #339, opened 2021-06-21
122. Best Time to Buy and Sell Stock II


ErdemT09:
https://leetcode.com/problems/best-time-to-buy-and-sell-stock-ii/

Varsayılan kod:
```java
class Solution {
    public int maxProfit(int[] prices) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/63192680/122808921-21fd1300-d2d6-11eb-9314-34f9147c1a94.png)
![image](https://user-images.githubusercontent.com/63192680/122808949-29242100-d2d6-11eb-9784-d453c72247fd.png)
![image](https://user-images.githubusercontent.com/63192680/122808981-3214f280-d2d6-11eb-930c-3182bf19df88.png)

PR #340, opened 2021-06-21
122. Best Time to Buy and Sell Stock II

ErdemT09:
Resolves: #339 

## Algorithm:

What we must do here is buy some stock and then sell it at the next price, if that price is not smaller than its next. This means that we can think of this as selling at individual prices and buying there again. 

For example: `1 3 4 9 18 5 6` -> buying at 1, then selling at 3 and buying their again, then selling at 4 and buying there agian and so on would be equivalent to buying at 1 and selling 18 because the prices is always increasing. When the price stops increasing, we don't buy the partial profit. At the last price of 6, we do buy because it is profitable, yielding us a total profit of `(18-1)+(6-5)`.

### DIFF
--- a/src/main/java/algorithms/easy/BestTimeToBuyAndSellStock2.java
+++ b/src/main/java/algorithms/easy/BestTimeToBuyAndSellStock2.java
@@ -1,0 +1,25 @@
+package algorithms.easy;
+
+public class BestTimeToBuyAndSellStock2 {
+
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Thank you for your valuable contribution Erdem.
Even if I don't like how they articulated this a little bit ambiguously, it is an important question as it is a basis for similar ones.

Approved.

Issue #341, opened 2021-06-22
309. Best Time to Buy and Sell Stock with Cooldown


ErdemT09:
https://leetcode.com/problems/best-time-to-buy-and-sell-stock-with-cooldown/

Varsayılan kod:
```java
class Solution {
    public int maxProfit(int[] prices) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/63192680/122865122-7d5aef80-d32e-11eb-9913-e8ebcaab05d0.png)
![image](https://user-images.githubusercontent.com/63192680/122865134-8350d080-d32e-11eb-9db0-0c3dcc918de7.png)

PR #342, opened 2021-06-22
309. Best Time to Buy and Sell Stock With Cooldown

ErdemT09:
Resolves: #341 

## Algorithm:
In every price, we make 3 decisions: 
- If we are in rest, out of the cooldown, we can buy here. 
- If we have bought, we can sell here. 
- If we are in rest, we can continue to be in rest.

It could be considered that we initially start at the rest state since there aren't any cooldowns limiting us. Buying a stock would mean subtracting a price from the rest profit. Selling a stock would mean adding the price of the stock to the current buying costs. 

We update the selling profit if it is more profitable to sell at this price than another price. We update the buying costs if it is less costly to buy here than a previous cost. The logic from #340 is also valid here. By some prices like `1 3 9`, to buy at 1 and sell at 9 is equivalent to buy at 1 and sell at 3, then buy at 3 and sell at 9.

The partial profit that the rest profit represents is not the total profit, which we inquire from selling here, rather it is the total profit before we have reassigned the total profit. The rest state would then represent some sort of partial profit, which is actually our total profit at the previous price.

### DIFF
--- a/src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithCooldown.java
+++ b/src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithCooldown.java
@@ -7,8 +7,12 @@ public class BestTimeToBuyAndSellStockWithCooldown {
         int totalProfit = 0, restProfit = 0, currBuyCosts = Integer.MIN_VALUE;
         for (int p : prices) {
             int prevBest = totalProfit;
-            totalProfit = Math.max(totalProfit, currBuyCosts + p);
-            currBuyCosts = Math.max(currBuyCosts, restProfit - p);
+            
+            if (currBuyCosts + p > totalProfit) {
+                totalProfit = p + currBuyCosts;
+            } else if (restProfit - p > currBuyCosts) {
+                currBuyCosts = restProfit - p;
+            }
             restProfit = Math.max(restProfit, prevBest);
         }
         return totalProfit;
--- a/src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithCooldown.java
+++ b/src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithCooldown.java
@@ -4,15 +4,15 @@ public class BestTimeToBuyAndSellStockWithCooldown {
 
     public int maxProfit(int[] prices) {
         
-        int maxProfit = 0, rest = 0, currBuyCost = Integer.MIN_VALUE;
+        int totalProfit = 0, rest = 0, currBuyCost = Integer.MIN_VALUE;
         for(int p : prices)
         {
-            int prevBest = maxProfit;
-            maxProfit = Math.max(maxProfit, currBuyCost+p);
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithCooldown.java
```diff
@@ -0,0 +1,24 @@
+package algorithms.medium;
+
+public class BestTimeToBuyAndSellStockWithCooldown {
+
+    public int maxProfit(int[] prices) {
+
+        int totalProfit = 0, restProfit = 0, currBuyCosts = Integer.MIN_VALUE;
```
ErdemT09, line 7: @altayhunoglu Do you know a better naming for the variable `currBuyCosts`. From finance, I can only think of "deficit" and "loss". But these only describe states where total profit is less than 0, not when are total profit goes below its current state because we have spent it on something, we might be still profiting here when we buy a stock, even before we sell.

ErdemT09, line 7: Is there a better naming for the variable `currBuyCosts`? It basically represents the state of our income when we have invested but have not earned from its returns.

altayhunoglu: How about investmentCost? 
"Cost of Investment" is also used commonly in finance.

altayhunoglu: It is called "`hold`" state in investment by the way.
Bitcoiners call it "`hodl`" as a joke, because of a Reddit post.

altayhunoglu:
That kind of table may help the future audience:
 
image


I just added some print statements into the code:
```java
class Solution {
 
    public int maxProfit(int[] prices) {
         System.out.println("  ACTIONS      | " + "totalProfit  |" + " restProfit   |" + " currBuyCosts");
        System.out.println("______________________________________________________");
         
        int totalProfit = 0, restProfit = 0, currBuyCosts = Integer.MIN_VALUE;
          System.out.println("INITIAL VALUES |  " + totalProfit + "              " + restProfit + "               " + currBuyCosts);
        for (int p : prices) {
            int prevBest = totalProfit;
            
            if (currBuyCosts + p > totalProfit) {
                totalProfit = p + currBuyCosts;
                 System.out.print("IF SELL AT : " + p);

            } else if (restProfit - p > currBuyCosts) {
                 currBuyCosts = restProfit - p;
                 System.out.print("IF BUY AT  : " + p);
            } 
             System.out.print(" |  " + totalProfit + "              " + restProfit + "               " + currBuyCosts);
             System.out.println("");
            restProfit = Math.max(restProfit, prevBest);
        }
        return totalProfit;
    }
}
```

altayhunoglu (approved):
Thank you Erdem.
Sorry for the late review.

Issue #343, opened 2021-06-22
714. Best Time to Buy and Sell Stock With Transaction Fee


ErdemT09:
https://leetcode.com/problems/best-time-to-buy-and-sell-stock-with-transaction-fee/
Vasrsayılan kod:

```java
class Solution {
    public int maxProfit(int[] prices, int fee) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/63192680/122872722-84d3c600-d339-11eb-99ba-e5317ca9fd9e.png)
![image](https://user-images.githubusercontent.com/63192680/122872794-a16ffe00-d339-11eb-9f68-e5ef06d640ec.png)

PR #344, opened 2021-06-22
714. Best Time to Buy and Sell Stock With Transaction Fee

ErdemT09:
Resolves: #343 

## Algorithm:
When we buy a stock, we are losing money from our total profits with the price of the stock. When we are selling at some price, we are not only adding the price of this stock back, but also losing money equivalent to the transaction fee. We should only update our total profit if selling with the consideration of the fee is more profitable than before. Transactions that have a profit of  0 or less when the fee is removed will not be taken into consideration as our total profit is initially 0.

### DIFF
--- a/src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithFee.java
+++ b/src/main/java/algorithms/medium/BestTimeToBuyAndSellStockWithFee.java
@@ -4,10 +4,14 @@ public class BestTimeToBuyAndSellStockWithFee {
 
     public int maxProfit(int[] prices, int fee) {
 
-        int totalProfit = 0, currBuyCosts = Integer.MIN_VALUE;
+        int totalProfit = 0, currBuyCosts = -1000000;
         for (int p : prices) {
-            totalProfit = Math.max(totalProfit, currBuyCosts+p-fee);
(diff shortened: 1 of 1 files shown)


altayhunoglu:
So, basically, we can interpret that question as a variant of https://github.com/spiralgo/algorithms/pull/340, but because of the fee, this quote is not valid for that question, right?

`For example: 1 3 4 9 18 5 6 -> buying at 1, then selling at 3 and buying their again, then selling at 4 and buying there agian and so on would be equivalent to buying at 1 and selling 18 because the prices is always increasing.`

ErdemT09:
> So, basically, we can interpret that question as a variant of #340, but because of the fee, this quote (by you) is not valid for that question, right?
> 
> `For example: 1 3 4 9 18 5 6 -> buying at 1, then selling at 3 and buying their again, then selling at 4 and buying there agian and so on would be equivalent to buying at 1 and selling 18 because the prices is always increasing.`

It's still equivalent, it's like we are not paying fees for those "mid-transactions". But it is not guaranteed if such an incremental thinking would be profitable when fees are added in.

altayhunoglu (approved):
Thank you for the collaboration Erdem.

PR #345, opened 2021-06-22
1197. Minimum Knight Moves

ErdemT09:
Resolves: #84 

## Algorithm:
To understand this problem, let us consider how we can reach to the target position:
### Approach 1-DFS Backwards:
The target position can be in either of 4 quadrants of the coordinate-chess plane. It's easier to only consider the case where it is in the first quadrant with both x and y being positive. Implementing this by taking their absolute value is sufficient.
We can think backwards as we can reach the starting position from the target position as well.  We can trace our path back to the starting position by going in either of the movements `(|X-2|,|Y-1|)`, `(|X-1|,|Y-2|)`. The absolute values are required as the positions can overflow to the negative quadrants of the plane. In such cases, we roll back to the positive. The minimum number of moves then becomes the minimum of the two tiles we can reach with these 2 moves plus 1. We then recursively follow from these all the way to the base cases that are the origin and the tiles `(1,1), (2,0), (0,2)`. Sometimes, we will naturally come across the same tiles again. We can store the values of the tiles that we reach for the first time and use this as means of memoizations, otherwise we would just visit too many tiles, making our algorithm unusably inefficient.

### Approach 2-BFS Forwards:
We can just start from the initial origin position and make our way to the target by applying the moves. The first time we reach it would yield the distance.

### Approach 3-Deriving a Mathematical Formula:
image
Looking at the grid, we can see a pattern emerging. At some places, the numbers such as 2 and 3 or 4 and 5 are interchanging, whereas at some point over the diagonal, the next interchanging group is appearing. We can eventually reduce this down to the formula in the code.

Image source: http://www.mathrecreation.com/2011/03/knight-moves.html

### DIFF
--- a/src/main/java/algorithms/curated170/medium/minimumknightmoves/MinimumKnightMoves.java
+++ b/src/main/java/algorithms/curated170/medium/minimumknightmoves/MinimumKnightMoves.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.medium;
+package algorithms.curated170.medium.minimumknightmoves;
 
 public class MinimumKnightMoves {
 
--- a/src/main/java/algorithms/curated170/medium/minimumknightmoves/MinimumKnightMovesBFS.java
+++ b/src/main/java/algorithms/curated170/medium/minimumknightmoves/MinimumKnightMovesBFS.java
@@ -5,7 +5,7 @@ import java.util.Deque;
 
 public class MinimumKnightMovesBFS {
 
-    private final int[][] DIRECTIONS = new int[][] { { 2, 1 }, { 1, 2 }, { -1, 2 }, { -2, 1 }, { -2, -1 }, { -1, -2 },
+    private final int[][] MOVES = new int[][] { { 2, 1 }, { 1, 2 }, { -1, 2 }, { -2, 1 }, { -2, -1 }, { -1, -2 },
             { 1, -2 }, { 2, -1 } };
 
     boolean[][] visited;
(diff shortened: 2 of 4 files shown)


altayhunoglu:
A classic question to teach Backtracking. Very important.

src/main/java/algorithms/curated170/medium/minimumknightmoves/MinimumKnightMoves.java
```diff
@@ -0,0 +1,29 @@
...
+        if (x == 2 && y == 2) {
+            return 4;
+        }
+        int delta = x - y;
+        if (2 * y > x) {
+            return (int) (delta - 2 * Math.floor((float) (delta - y) / 3));
+        } else {
+            return (int) (delta - 2 * Math.floor((float) (delta - y) / 4));
+        }
+    }
```
altayhunoglu, line 27: I think, this is really a good explanation how we derive this formula:

image

https://stackoverflow.com/questions/2339101/knights-shortest-path-on-chessboard/41704071#comment74156588_41704071

src/main/java/algorithms/curated170/medium/minimumknightmoves/MinimumKnightMovesBFS.java
```diff
@@ -0,0 +1,56 @@
+package algorithms.curated170.medium.minimumknightmoves;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+public class MinimumKnightMovesBFS {
```
altayhunoglu, line 6: BFS solution reminded me of the biosonar that bats use to measure distance:
https://www.youtube.com/watch?v=laeE4icRYp4

altayhunoglu (approved):
Danke für die gute Zusammenarbeit.
Der Pull-Request wird genehmigt.

PR #346, opened 2021-07-01
333. Largest BST Subtree

ErdemT09:
Resolves: #82 

## Algorithm: 
We need to first recall what it means to have a tree that qualifies to be a binary search tree. For a simple recursive definition, a binary search tree has its root being parent of two binary search tree roots and this root's value is larger* than all values in the tree of its left child and smaller* than all the values in the tree of its right child.

If these conditions are satisfied, the size of the BST here is the sizes of the left subtree and the right subtree + 1 (the root itself). If the conditions are not met, otherwise, it would be the maximum of the either two subtree, whether they themselves are  binary search tree or they simply contain one of that size. 

Then we simply need to recurse our way up through the leaves to the root node, bringing 3 values with us:
- The minimum value in this subtree
- The maximum value in this subtree
- The maximum size of a binary search tree within this subtree

Additionally, if a subtree is not a BST, then we can do a simply inversion to have all the parent subtrees be invalid: We can set the minimum value to positive infinity and the maximum value to negative infinity.

*or in some cases also equal, but we consider such a case invalid here

### DIFF
--- a/src/main/java/algorithms/curated170/medium/LargestBSTSubtree.java
+++ b/src/main/java/algorithms/curated170/medium/LargestBSTSubtree.java
@@ -18,11 +18,15 @@ public class LargestBSTSubtree {
         }
         int[] left = largestBST(node.left);
         int[] right = largestBST(node.right);
-        if (node.val > left[MAX_VAL] && node.val < right[MIN_VAL]) {
+        if (isValidBST(node, left, right)) {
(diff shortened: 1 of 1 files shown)


altayhunoglu:
For this case the answer is two, because of these nodes, right?
image

ErdemT09:
> For this case the answer is two, because of these nodes, right?

1-2 subtree on the bottom left is also valid.

altayhunoglu (approved):

PR #347, opened 2021-07-03
545. Boundary of Binary Tree

ErdemT09:
Resolves: #79 

## Algorithm:
As stated in the problem, an algorithm consisting of 3 procedures is required to solve this problem:
### 1. Get the Left Boundary of the Tree:
Here, we go as left as we can. If we encounter a node without left children, we go to its right and go as left as we can from there, until we hit the leaf with no roots, which is not considered to be part of the left boundary, belonging to the leaves group instead. 
### 2. Get the Right Boundary of the Tree
We basically apply the same procedure as the left boundary one. We go right as much as we can and turn left if necessary, avoiding the right-most leaf at the end.
### 3. Get the Leaves of the Tree
For this, we apply a more traditional DFS algorithm (other ones are DFS too, but this is more clear in this regard). We go down through each node to their right and left. If we come across a node without children, we add it to our boundary list.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/BoundaryOfBinaryTree.java
+++ b/src/main/java/algorithms/curated170/medium/BoundaryOfBinaryTree.java
@@ -1,0 +1,59 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import algorithms.datastructures.TreeNode;
+
+public class BoundaryOfBinaryTree {
+    public List<Integer> boundaryOfBinaryTree(TreeNode root) {
+        List<Integer> res = new ArrayList<>();
+        if (root == null) {
+            return res;
+        }
+        res.add(root.val);
+
+        printLeftBorder(root.left, res);
+        if (root.left != null || root.right != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/medium/BoundaryOfBinaryTree.java
```diff
@@ -0,0 +1,59 @@
...
+        if (root == null) {
+            return boundary;
+        }
+        boundary.add(root.val);
+
+        getLeftBorder(root.left, boundary);
+        if (root.left != null || root.right != null) {
+            getLeaves(root, boundary);
+        }
+        getRightBorder(root.right, boundary);
```
altayhunoglu, line 20:
```suggestion
         if (root.left != null || root.right != null) {
             getLeftBorder(root.left, boundary);
             getLeaves(root, boundary);
             getRightBorder(root.right, boundary);
        }
```

image

altayhunoglu (approved):
Thanks for the collaboration Erdem.
Very nice, easy-to-understand coding style.

Issue #169, opened 2021-05-06
Theory: Can all algorithm questions that can be solved using either BFS or DFS be solved using the other one?


ErdemT09:
The question is in the title. 
This question appeared to me when the same question ([here](https://github.com/spiralgo/algorithms/tree/master/src/main/java/algorithms/curated170/medium/killprocess)) were using BFS and DFS. Is such a principle that if a problem is solvable by one yields the fact that it is also solvable by the another universally valid?

altayhunoglu:
Nice question Erdem.
 
 I am not sure but, maybe we can reduce this question like this:
 
 - If a problem can be solved using an iterative way (for, while); can it also be solved using a recursive way?
  Or maybe like this:
 - If a problem can be solved using a Queue; can it also be solved using a Stack?

  I have to meditate and investigate more about this question, in order to provide a certain answer.
  
  But currently, I think; it is possible. But some problems are solved easier when you think backwards (recursively) instead of forwards (iteratively)
  
   I recognised these when I tried to solve "Tower of Hanoi" using an iterative way instead of the classic recursive solution.
   I found it very hard to solve iteratively.

ErdemT09:
Here is what Wikipedia says on [its page about DFS](https://en.wikipedia.org/wiki/Depth-first_search):
> The non-recursive implementation is similar to breadth-first search but differs from it in two ways:
> 
> 1. it uses a stack instead of a queue, and
> 2. it delays checking whether a vertex has been discovered until the vertex is popped from the stack rather than making this check before adding the vertex.
> If G is a tree, replacing the queue of the breadth-first search algorithm with a stack will yield a depth-first search algorithm. For general graphs, replacing the stack of the iterative depth-first search implementation with a queue would also produce a breadth-first search algorithm, although a somewhat nonstandard one.

PR #348, opened 2021-07-03
The Ackermann Function

ErdemT09:
Ackermann's function is an interesting example of how complex computations can emerge from simple functional rules. When two numbers are inputted into this function, there are 3 cases:

1. The first one is 0, we then return the second number +1.
2. The second one is 0, we then return ack of the first number minus 1, and reducing the second number to 1 in the input.
3. Otherwise, we do what one might call a "nested recursion".  This can sound like a pleonasm, as all recursions are inherently nested. However, here, we first calculate the value of the function for some input values, then we take this return value and input it to our function as the secondary value. 

[Here](https://en.wikipedia.org/wiki/Ackermann_function#Example_expansions), it can be seen how uncontrolled this functional behavior grows.
It's also important to note that such a function is described to be `not primitive recursive`, where `primitive recursive function` can be nonrigorously defined as any recursive function that can be represented by a for- or while-loop.


altayhunoglu:
Thank you for bringing this subject here Erdem.
 It is somehow related to the question that you had asked once: https://github.com/spiralgo/algorithms/issues/169
 
 This video is also quite informative on this issue:
https://www.youtube.com/watch?v=i7sm9dzFtEI&ab_channel=Computerphile

altayhunoglu (approved):
Thank you for bringing this subject here Erdem.
 It is somehow related to the question that you had asked once: https://github.com/spiralgo/algorithms/issues/169
 
 This video is also quite informative on this issue:
https://www.youtube.com/watch?v=i7sm9dzFtEI&ab_channel=Computerphile

PR #349, opened 2021-07-03
353. Design Snake Game

ErdemT09:
Resolves: #85 

## Algorithm:

The thing that we can notice is that the snake is just a linked list of points on the grid that we are playing the game on. Using this intuition, we can easily represent the snake using a linked list of points, for which we can create a simple auxiliary class.

To move in any given input direction, we can peek the head of the snake and increment it in the given direction. If this new head of the snake intersects any of the points of the snake besides its moving tail, this means that the game should be over. 
If this new head tried to move outside the bounds of the grid, this means that the game should be over.

Since all food does not appear on the screen at once, rather one-by-one, we can keep track of which food is on the screen through the score. If we the new head is in some food tile, then we increment the score counter and return the new score, with the incremented score also operating to make the new food "appear" on the grid.
If snake doesn't eat food, that means its length stays constant and its tail should be removed.
After doing all these, we return the score if possible.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/DesignSnakeGame.java
+++ b/src/main/java/algorithms/curated170/medium/DesignSnakeGame.java
@@ -46,13 +46,11 @@ public class DesignSnakeGame {
         }
 
         private boolean snakeEatsFood(Point newHead) {
-            return (score < FOOD.length && FOOD[score][0] == newHead.X 
-                    && FOOD[score][1] == newHead.Y);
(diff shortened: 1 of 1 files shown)


altayhunoglu:
Very nice problem to grasp the basics of object-oriented programming.
LinkedList is a good choice for this context, indeed.

altayhunoglu (approved):
Very nice problem to grasp the basics of object-oriented programming.
LinkedList is a good choice for this context, indeed.

Very nice, clean, and modular coding style. Thank you, Erdem.

PR #350, opened 2021-07-03
271. Encode and Decode Strings

ErdemT09:
Resolves: #87 

## Algorithm:
The most straightforward way of addressing this problem is by putting delimiters between the strings in the given list and then unpacking the encoded string with consideration of these delimiters.
There are 3 ways of implementing  this in Java:

### 1. Combine the Strings by Adding the String Length and a Delimiter Before a String
When encoding, we can append the string length of the string in front of it, then a delimiter such as `'/'` to distinguish the length from the actual string, then we can append the string itself after these. We can easily combine all the strings from the list in such a way.
Decoding is the actual noteworthy procedure here. The delimiter is located somewhere in the string after the end of the last string, between these is the length of the string. So we can find where the delimiter is and then get the substring therebetween as the integer value representing the length of the next string. We can then get this string that starts from the index after the delimiter and adds at the location corresponding to the length we have found. 

### 2. Combine the Strings by Adding the String Length Itself as a Delimiter
Thinking about strings, they are actually arrays of base-256 numbers. So we can just append the length of the string as a character itself. Sometimes, this character may not even be visible when printed. A string of length 97 would have the character `'a'` appended in front of it. We can decode the string in a similar manner as we have done above, without using any extra delimiters.

### 3. Replace and Append a Delimiter + Streams
Just appending delimiters to the end or the beginning of the string is not possible as the strings themselves may contain such delimiters. We can then replace the delimiters within the strings with a duplicate of them to distinguish and append the delimiter with spaces before and after it to distinguish further. 
To decode, we can split the string into such parts and re-replace the old present delimiters.
This one is a bit slower, because of character replacements and the streams.

### DIFF
--- a/src/main/java/algorithms/curated170/medium/encodeanddecodestrings/EncodeAndDecodeStrings.java
+++ b/src/main/java/algorithms/curated170/medium/encodeanddecodestrings/EncodeAndDecodeStrings.java
@@ -1,0 +1,38 @@
+package algorithms.curated170.medium;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EncodeAndDecodeStrings {
+
+    public class Codec {
+
+        public String encode(List<String> strs) {
+            StringBuilder encoder = new StringBuilder();
+            strs.forEach(s -> encoder.append((char) s.length()).append(s));
+            return encoder.toString();
+        }
+
+        public List<String> decode(String s) {
+            List<String> decodedStrs = new ArrayList<>();
+            short i = 0, len;
+
+            while (i < s.length()) {
+                len = (short) (s.charAt(i) + i + 1);
+                decodedStrs.add(s.substring(i + 1, len));
+                i = len;
+            }
+            
+            return decodedStrs;
+        }
+    }
+
+    public static void main(String[] args) {
+        var solution = new EncodeAndDecodeStrings();
+        Codec codec = solution.new Codec();
+        String en = codec.encode(List.of("One", "Two", "Three", "Four", "", " ",
+                "averylongstringaverylongstringaverylongstringaverylongstring"));
+        System.out.println(en);
+        System.out.println(codec.decode(en));
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/algorithms/curated170/medium/encodeanddecodestrings/EncodeAndDecodeStringsSlashDelimiter.java
```diff
@@ -0,0 +1,39 @@
+package algorithms.curated170.medium.encodeanddecodestrings;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EncodeAndDecodeStringsSlashDelimiter {
+
```
altayhunoglu, line 7: Test case: `["Hello","World"]`
Encoded String:  `"5/Hello5/World"`

src/main/java/algorithms/curated170/medium/encodeanddecodestrings/EncodeAndDecodeStrings.java
```diff
@@ -0,0 +1,38 @@
+package algorithms.curated170.medium.encodeanddecodestrings;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EncodeAndDecodeStrings {
+
```
altayhunoglu, line 7: Test case: `["Hello","World"]`
Encoded String:  `"\u0005Hello\u0005World"`

src/main/java/algorithms/curated170/medium/encodeanddecodestrings/EncodeAndDecodeStringsStream.java
```diff
@@ -0,0 +1,21 @@
+package algorithms.curated170.medium.encodeanddecodestrings;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class EncodeAndDecodeStringsStream {
+
```
altayhunoglu, line 8: Test case: `["Hello","World"]`
Encoded String:  `"Hello # World # "`

altayhunoglu (approved):
Thank you for your valuable efforts, Erdem.

PR #351, opened 2021-07-03
708. Insert into a Sorted Circular Linked List

ErdemT09:
Resolves: #88 

## Algorithm:
Here, it's best to think about what might occur when we are inserting the value into such a list:
It could be inserted between values. 5 will naturally be inserted after 4 and before 6.
It could be inserted at the end of the list. At the end of the list, the next value is smaller than the previous one. A node would be inserted here if and only if succeeds the largest value or precedes the smallest value. 
The algorithm follows from these simple observations.


altayhunoglu (approved):
A nice explanation, easy-to-understand solution, and readable code. Thanks, Erdem.

PR #352, opened 2021-07-03
431. Encode N-ary Tree to Binary Tree

ErdemT09:
Resolves: #89 

## Algorithm:
There are few ways of implementing this. They are all inherently similar as they try to convert the parent-child relationship from the n-ary tree into a left-right relation of the the binary tree.
We can interpret all the children of the root node listed to its left like a linked list. In each of these nodes, we can interpret their children being listed towards the other direction, right, as a linked list. For their children the direction of listing becomes left. So we can just go left and then right to turn the n-ary tree into a combination of linked lists represented by the binary tree.
image


altayhunoglu:
Nice images Erdem. Thank you.

altayhunoglu (approved):
Very beautiful encoding strategy, indeed.
Thanks, Erdem. Nothing I can add.

PR #353, opened 2021-07-03
489. Robot Room Cleaner 

ErdemT09:
Resolves: #90 

## Algorithm:
Here, it sounds pretty natural to employ some kind of backtracking. If we can't move in one direction, we would naturally turn the robot to some other direction. If the robot moves in some direction beginning from some tile, to move it to right from that tile, we need to backtrack the robot at each step it has moved. We should also keep a record of which tiles we've already cleaned in order to not visit and move from them again.
This is of course corresponding with general template of backtracking at #304.


altayhunoglu (approved):
Another nice implementation of Backtracking.
I think, this code can be also applied for the real robots.

PR #354, opened 2021-07-03
759. Employee Free Time

ErdemT09:
Resolves: #91 

## Algorithm:
The point here is that any free time interval of the employees is equivalent to some time interval beginning at the end of one employee's time and ending at the other's start, where the start is later the end of the previous interval. We can then just iterate over the intervals sorted by their starting time, and add new free time intervals if the current start is later than the latest end time we have encountered. For recording the latest end time, we would simply need an integer. To iterate over the intervals in such a sorted manner, we can either directly sort it or use a priority queue. 

One can also revisit #64 in order to perhaps get a view of this problem from another point.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/employeefreetime/EmployeeFreeTimePriorityQueue.java
+++ b/src/main/java/algorithms/curated170/hard/employeefreetime/EmployeeFreeTimePriorityQueue.java
@@ -1,0 +1,42 @@
+package algorithms.curated170.hard.employeefreetime;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.PriorityQueue;
+
+public class EmployeeFreeTimePriorityQueue {
+
+    public List<Interval> employeeFreeTime(List<List<Interval>> schedule) {
+        PriorityQueue<Interval> earliestIntervals = new PriorityQueue<Interval>((a, b) -> a.start - b.start);
+        schedule.forEach(e -> earliestIntervals.addAll(e));
+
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/hard/employeefreetime/EmployeeFreeTimePriorityQueue.java
```diff
@@ -0,0 +1,42 @@
...
+        PriorityQueue<Interval> earliestIntervals = new PriorityQueue<Interval>((a, b) -> a.start - b.start);
+        schedule.forEach(e -> earliestIntervals.addAll(e));
+
+        int end = earliestIntervals.poll().end;
+        List<Interval> commonIntervals = new ArrayList<>();
+
+        while (!earliestIntervals.isEmpty()) {
+            Interval interval = earliestIntervals.poll();
+
+            if (interval.start > end && end >= 0) {
```
altayhunoglu, line 19:
```suggestion
            if (interval.start > end) {
```
Constraint:
`0 <= schedule[i].start < schedule[i].end <= 10^8
`

ErdemT09: It's a mistake. As in the other file, end is initialized as -1.

altayhunoglu (approved):
Thank you for your all attention and help Erdem.

altayhunoglu:
It is how we got the result: `[[5,6],[7,9]]`

IMG-1707

PR #355, opened 2021-07-04
1274. Number of Ships in a Rectangle

ErdemT09:
Resolves: #92 

## Algorithm:
Here, we utilize a kind of divide-and-conquer algorithm. We can always split the given rectangle into 4 parts and count the number of ships in each one of them. This naturally implies recursion. The base cases of our counting are that we find a rectangle without ships or we find a rectangle that is actually only point on the grid, which suggests that there can only be a single ship here. We combine all these cases to count the total amount for the input rectangle.


altayhunoglu (approved):
What a brilliant solution. Simple, yet elegant.
Thanks Erdem.

Issue #219, opened 2021-05-15
302. Smallest Rectangle Enclosing Black Pixels


altayhunoglu:
https://leetcode.com/problems/smallest-rectangle-enclosing-black-pixels/
Varsayılan kod:

```
class Solution {
    public int minArea(char[][] image, int x, int y) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380511-edb57900-b5ea-11eb-879d-0aee5238b7ac.png)
![image](https://user-images.githubusercontent.com/60903744/118380514-f443f080-b5ea-11eb-8478-0a281881158a.png)
![image](https://user-images.githubusercontent.com/60903744/118380517-fa39d180-b5ea-11eb-894d-a74090f8e650.png)

PR #356, opened 2021-07-05
302. Smallest Rectangle Enclosing Black Pixels

ErdemT09:
Resolves: #219 
Recommended problems before this:
[200. Number of Islands](https://github.com/spiralgo/algorithms/blob/master/src/main/java/algorithms/medium/NumberOfIslands.java) : https://leetcode.com/problems/number-of-islands/
1292. Maximum Side Length of a Square with Sum Less than or Equal to Threshold: #358 
1765. Map of Highest Peak: #357 

## Algorithm:
The smallest rectangle area would correspond to a reactangle covering the left- and right-most points of the black pixels as well as the top and the bottom boundaries.
### Approach 1-Binary Search
The intuition here is that the black pixel at the bottom is located somewhere between the given x-coordinate (coordinates for 2D-arrays aren't defined properly) and 0. If we find a black pixel in any of the pixels in the row between these, that means that we must look further down to see if there is a row with a black pixel in it. Else, we look between the middle and the x. This naturally corresponds to binary-searching all the rows between these. For the upper case, we always try to look for the upper rows. For left and right, our boundaries are not just the matrix boundaries, but the already calculated bottom and top values. While searching columns, we don't look for the pixels in those areas.
### Approach 2-DFS
This approach is more straight forward but is slower, especially in cases where the majority of the image is black pixels. Since all the black pixels are connected, we can just search through the given pixel to find the boundaries. We can also mark visited nodes to `'1'` to save that they are visited.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/smallestrectangleenclosingblackpixels/SmallestRectangleEnclosingBlackPixelsBinarySearch.java
+++ b/src/main/java/algorithms/curated170/hard/smallestrectangleenclosingblackpixels/SmallestRectangleEnclosingBlackPixelsBinarySearch.java
@@ -1,0 +1,38 @@
+package algorithms.curated170.hard.employeefreetime;
+
+public class SmallestRectangleEnclosingBlackPixelsDFS {
+
+    private char[][] image;
+
+    public int minArea(char[][] image, int x, int y) {
+        this.image = image;
+        int m = image.length, n = image[0].length;
+
+        int bottom = searchBoundary(0, x, 0, n, true, true);
+        int top = searchBoundary(x + 1, m, 0, n, false, true);
+        int left = searchBoundary(0, y, bottom, top, true, false);
+        int right = searchBoundary(y + 1, n, bottom, top, false, false);
+
+        return (right - left) * (top - bottom);
+    }
+
+    private int searchBoundary(int i, int j, int low, int high, boolean findMin, boolean isRow) {
+        while (i != j) {
+            int k = low, mid = (i + j) / 2;
+            while (k < high && isBlack(mid, k, isRow)) {
+                k++;
+            }
+            if ((k < high) == findMin) {
+                j = mid;
+            } else {
+                i = mid + 1;
+            }
+        }
+
+        return i;
+    }
+
+    private boolean isBlack(int mid, int k, boolean isRow) {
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/hard/smallestrectangleenclosingblackpixels/SmallestRectangleEnclosingBlackPixelsDFS.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        if (y < bottom) {
+            bottom = y;
+        }
+        if (y > top) {
+            top = y;
+        }
+        if (x > right) {
+            right = x;
+        }
```
altayhunoglu, line 40: Although programmatically it is not a problem, it would be better to keep the coordinate notation right:
  ```suggestion
   if (y < left) {
            left = y;
        }
        if (x < bottom) {
            bottom = x;
        }
        if (x > top) {
            top = x;
        }
        if (y > right) {
            right = y;
        }

```

File: src/main/java/algorithms/curated170/hard/smallestrectangleenclosingblackpixels/SmallestRectangleEnclosingBlackPixelsBinarySearch.java [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+        return (right - left) * (top - bottom);
+    }
+
+    private int searchBoundary(int i, int j, int low, int high, boolean findMin, boolean isRow) {
+        while (i != j) {
+            int k = low, mid = (i + j) / 2;
+            while (k < high && isBlack(mid, k, isRow)) {
+                k++;
+            }
+            if ((k < high) == findMin) {
```
altayhunoglu, line 25: findMin is quite a smart idea to generalize the Binary Search method for both rows and cols.

src/main/java/algorithms/curated170/hard/smallestrectangleenclosingblackpixels/SmallestRectangleEnclosingBlackPixelsDFS.java
ErdemT09: I think considering x to be vertical and y to be horizontal is less suitable.

altayhunoglu (approved):
 Approved.

PR #359, opened 2021-07-05
198. House Robber & 213. House Robber II

ErdemT09:
## [198. House Robber](https://leetcode.com/problems/house-robber/)
Here, the maximum robbed money at every house would equal to the maximum of: the amount stolen at the previous house, the amount stolen at the house before + the current house's money.
We can hence turn this problem into a dynamic programming problem, where the result is the amount stolen at the last house.
## [213. House Robber II](https://leetcode.com/problems/house-robber-ii/)
As the first and the last houses are connected, the maximum sum equals the maximum of the maximum sum of the houses without the first or without the last. We apply the algorithm from the first problem in two passes and compare the results.

### DIFF
--- a/src/main/java/algorithms/medium/HouseRobberII.java
+++ b/src/main/java/algorithms/medium/HouseRobberII.java
@@ -3,9 +3,13 @@ package algorithms.medium;
 public class HouseRobberII {
 
     public int rob(int[] nums) {
+        int len = nums.length;
+        if (len == 1) {
+            return nums[0];
+        }
 
         int nowStolenA = 0, prevStolen = 0, stolen = 0;
-        for (int i = 0; i < nums.length - 1; i++) {
+        for (int i = 0; i < len - 1; i++) {
             prevStolen = nowStolenA;
             nowStolenA = Math.max(prevStolen, stolen + nums[i]);
             stolen = prevStolen;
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/medium/HouseRobberII.java
```diff
@@ -0,0 +1,25 @@
...
+        for (int i = 0; i < nums.length - 1; i++) {
+            prevStolen = nowStolenA;
+            nowStolenA = Math.max(prevStolen, stolen + nums[i]);
+            stolen = prevStolen;
+        }
+
+        prevStolen = 0;
+        stolen = 0;
+        int nowStolenB = 0;
+        for (int i = 0; i < nums.length; i++) {
```
altayhunoglu, line 17:
```suggestion
        for (int i = 1; i < nums.length; i++) {
```

src/main/java/algorithms/medium/HouseRobberII.java
```diff
@@ -0,0 +1,25 @@
+package algorithms.medium;
+
+public class HouseRobberII {
+
+    public int rob(int[] nums) {
+
```
altayhunoglu, line 6:
```suggestion
        if (nums.length == 0)
            return 0;

        if (nums.length == 1)
            return nums[0];
```

ErdemT09: We don't need the 0-check. The 1-check is needed though.

altayhunoglu (approved):
Thanks for the valuable contribution Erdem.

Issue #243, opened 2021-05-15
656. Coin Path


altayhunoglu:
https://leetcode.com/problems/coin-path/

Varsayılan kod:

```
class Solution {
    public List<Integer> cheapestJump(int[] coins, int maxJump) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118381013-438c2000-b5ef-11eb-8d1f-647305d0e329.png)
![image](https://user-images.githubusercontent.com/60903744/118381014-4850d400-b5ef-11eb-959b-f4b52ee5c1c1.png)

PR #360, opened 2021-07-05
656. Coin Path

ErdemT09:
Resolves: #243 

## Algorithm:
The naive way of doing this would be trying to jump from the start to every possible index, and from there to every possible index and so on. With this, it should be obvious that we might be calculating for each index more than once. Instead of this, we can apply dynamic programming here. 

Instead of trying to jump forward, we can work backwards. The cost of jumping to last index from the last index itself is the value there. At the index before it, the cost would be the value there would be the value at this index + the target index. So, at every index, we consider all indices that we could jump to this index from, pick the minimum and add it to our paths data. To compare the costs, we store the costs in the input array to compare them when needed. So, the result by induction is that we eventually calculate the minimum cost for the 0-index, with the aid of which we also calculate the path.


src/main/java/algorithms/curated170/hard/CoinPath.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class CoinPath {
+    public List<Integer> cheapestJump(int[] A, int B) {
+        int coins = A.length;
+        if (A == null || coins < 1 || A[coins - 1] < 0) {
+            return Collections.emptyList();
```
altayhunoglu, line 11:
```suggestion
            return List.of();
```
Maybe you prefer using the new and shorter way. 
There is no efficiency difference, but there is a subtle difference about serialization and exception handling that we needn't consider for this context.

src/main/java/algorithms/curated170/hard/CoinPath.java
```diff
@@ -0,0 +1,53 @@
...
+
+            if (minCost != Integer.MAX_VALUE) {
+                A[i] += minCost;
+            } else {
+                A[i] = -1;
+            }
+        }
+
+        if (A[0] == -1) {
+            return Collections.emptyList();
```
altayhunoglu, line 42:
```suggestion
            return List.of();
```
Suggestion for the same reason.

src/main/java/algorithms/curated170/hard/CoinPath.java
ErdemT09: I chose `emptyList()` because it is easier to read. I didn't look up its technical properties though.

ErdemT09: ıt seems that `emptyList()` actually does a lot in  the background while `List.of()` does simpler operations. I will follow your advice.

altayhunoglu (approved):
Thank you Erdem. Quite a readable code and nice solution.

Issue #212, opened 2021-05-15
296. Best Meeting Point


altayhunoglu:
https://leetcode.com/problems/best-meeting-point/

Varsayılan kod:

```
class Solution {
    public int minTotalDistance(int[][] grid) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380254-f60cb480-b5e8-11eb-9501-e02632606b31.png)
![image](https://user-images.githubusercontent.com/60903744/118380258-fdcc5900-b5e8-11eb-835f-9a55983f2c35.png)
![image](https://user-images.githubusercontent.com/60903744/118380260-0329a380-b5e9-11eb-9e97-f1a3d9bf9dd2.png)

PR #361, opened 2021-07-05
453. Minimum Moves to Equal Array Elements

ErdemT09:
Problem:
https://leetcode.com/problems/minimum-moves-to-equal-array-elements/

Here, incrementing all numbers except one is equivalent to decrementing that one number, as we only care about the fact that the numbers are equal. Thus, the solution is equivalent to decrementing every number to the minimum in the array. The amount of operations done to the each element would result in the general formula `sum - length*minimum`.

Array problems as such are good build-ups for this larger problem: #212

### DIFF
--- a/src/main/java/algorithms/easy/MinimumMovesToEqualArrayElements.java
+++ b/src/main/java/algorithms/easy/MinimumMovesToEqualArrayElements.java
@@ -1,0 +1,14 @@
+package algorithms.easy;
+
+public class SingleNumber {
+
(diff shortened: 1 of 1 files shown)


ErdemT09:
Problem for the follow-up: https://leetcode.com/problems/minimum-moves-to-equal-array-elements-ii
How can we find the median of an unsorted array in O(n) time?

altayhunoglu (approved):
Again, nice explanation Erdem. Thanks for the valuable contribution.

PR #362, opened 2021-07-06
Quickselect Algorithm

ErdemT09:
Qucikselect is an algorithm used for finding the kth smallest element in an unsorted list. Its procedure of doing this contains similarites to the quicksort algorithm, they are both developed by Tony Hoare.
In an array, it's rather trivial to find the maximum or the minimum. But things get hard if we are trying to find the median or the 12th smallest element out of 100. We can of course sort, but that would take linearithmic time. Quickselect performs in linear time on avarage, it partially sorts the array as well. 
To find the kth smallest element, we choose some random pivot. We then transfer every element smaller than this pivot the the start of the array. If the transfer index, new pivot, is greater than k, that means that the kth element is among these. If the new pivot is larger, that means that the kth element is among the elements that were not transferred. So, we partition the array each time and eventually find the kth element.

It could also be implemented as an iterative algorithm, where we can also move right elements to the other side of the pivot.

### DIFF
--- a/src/main/java/algorithms/QuickSelect.java
+++ b/src/main/java/algorithms/QuickSelect.java
@@ -2,9 +2,9 @@ package algorithms;
 
 import java.util.Arrays;
 
-class Quickselect {
+public class Quickselect {
 
-  int select(int[] arr, int left, int right, int k) {
+  public int select(int[] arr, int left, int right, int k) {
     if (left == right) {
       return arr[left];
     }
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/QuickSelect.java
```diff
@@ -0,0 +1,73 @@
+package algorithms;
+
+import java.util.Arrays;
+
+public class Quickselect {
```
altayhunoglu, line 5:
```suggestion
public class QuickSelect{
```
In order to solve the error message.

ErdemT09: File's name is wrong.

Issue #241, opened 2021-05-15
269. Alien Dictionary


altayhunoglu:
https://leetcode.com/problems/alien-dictionary/

Varsayılan kod:

```
class Solution {
    public String alienOrder(String[] words) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380958-c6f94180-b5ee-11eb-8003-332d1c42f32d.png)
![image](https://user-images.githubusercontent.com/60903744/118380961-cb255f00-b5ee-11eb-9706-5a7802923ed7.png)
![image](https://user-images.githubusercontent.com/60903744/118380964-cf517c80-b5ee-11eb-95e1-c46ed02dd4f5.png)

PR #366, opened 2021-07-08
207. Course Schedule & 210. Course Schedule II

ErdemT09:
For this question, it is best to review the algorithms that were used to solve #180 and #190. 
## 207. Course Schedule
Problem: https://leetcode.com/problems/course-schedule/

The problem here is basically detecting a cycle in a directed graph. 
There are a few ways of approaching this but the most straightword ways are topological sorting and counting using BFS and checking if a vertex was visited before in DFS. 
In BFS-Topological Sort, if we end the loop when there are vertices with in-degree greater than 0, then we didn't the traverse the entire graph. We can keep the number of vertices that we poll and compare at the end.

## 210. Course Schedule II
Problem: https://leetcode.com/problems/course-schedule-ii/
Here, we need the actual topological sort order of the vertices of a digraph. We can once again apply Kahn's algorithm for this. It's simple as we only add the vertices we poll to a list and return it.
In DFS, topological sorting is achieved by working backwards. If I have a graph like `0->1->2`, by the logic of DFS, I am able to process the last vertex first. I should then add this vertex to the end of the list. It all works backward from there.

The algorithms discussed here will come to use in #241.

### DIFF
--- a/src/main/java/algorithms/medium/CourseSchedule.java
+++ b/src/main/java/algorithms/medium/CourseSchedule.java
@@ -1,0 +1,57 @@
+package algorithms.medium;
+
+public class CourseSchedule {
+
+    boolean[] visiting;
+    boolean[] checked;
+
+    public boolean canFinish(int numCourses, int[][] prerequisites) {
+        int[][] digraph = makeGraph(prerequisites, numCourses);
+        visiting = new boolean[numCourses];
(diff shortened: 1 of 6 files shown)


altayhunoglu (approved):
I could not see anything to add.
Both BFS and backtracking solutions are quite fine to me.
Thanks Erdem.

Issue #217, opened 2021-05-15
1231. Divide Chocolate


altayhunoglu:
https://leetcode.com/problems/divide-chocolate/

Varsayılan kod:
 
```
 class Solution {
    public int maximizeSweetness(int[] sweetness, int K) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380449-7253c780-b5ea-11eb-87d7-5123f0cd897d.png)
![image](https://user-images.githubusercontent.com/60903744/118380460-83043d80-b5ea-11eb-9897-bcfd7882f408.png)

PR #369, opened 2021-07-09
1231. Divide Chocolate

ErdemT09:
Resolves: #217 

## Algorithm:
I recommend looking at these problems #368, #311 and #122 before moving onto this problem.
Like in those problems, we can try to guess our return value. If we pick some "sweetness value" and amount of splits required for it are more than allowed, this means that it's too low and we need to reduce it to split more. If the number of splits is too high or equal, we can try a higher value. We can easily utilize binary searching for this.


altayhunoglu (approved):
Thank you erdem. A very nice application of Binary Search.

Issue #224, opened 2021-05-15
465. Optimal Account Balancing


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int minTransfers(int[][] transactions) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380602-ba271e80-b5eb-11eb-8757-88d6873dde73.png)
![image](https://user-images.githubusercontent.com/60903744/118380607-c27f5980-b5eb-11eb-8ead-09058c905238.png)

PR #370, opened 2021-07-09
465. Optimal Account Balancing

ErdemT09:
Resolves: #224 

## Algorithm:
If someone gives me $10 and I give that person $10 back, it would mean that there are 0 transactions needed. 
If I give that person only $8 back, the situation is equivalent to me getting only $2 from the person.
If someone comes and gives the person in $2 deficit that money, the situation is now equivalent to the new person giving me $2. So, we calculate how much each person is in deficit/surplus. If their accounts are balanced with $0 change, we don't include them in the debts list, as no transactions will be done with them. 
Some debts will directly cancel each other. In the state above, the debts would look like this: `[-2, 2]` (second person is balanced). These values would directly cancel each other, so we don't need extra computation for it. But what if the debts look like this: `[-4, -4, 2, 6]`. We can try paying the first person with the 3. Then we have [0, -4, -2, 6]. We can pay the second with the last to get [0, 0, -2, 2]. The last two can cancel each other out and we would be left with 3 transactions. This is in fact the solution here, but it may not be in other situations. So, we might have to try all opposite-signed balances. For this, to cancel out the side-effects of the previous trial, we can use simple backtracking.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/OptimalAccountBalancing.java
+++ b/src/main/java/algorithms/curated170/hard/OptimalAccountBalancing.java
@@ -1,86 +1,99 @@
 package algorithms.curated170.hard;
 
-public class OptimalAccountBalancing {
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 
+public class OptimalAccountBalancing {
+    
     public int minTransfers(int[][] transactions) {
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
This question really deserves the `hard` label, according to me. 
  Thanks Erdem. Nice solution.

Issue #220, opened 2021-05-15
425. Word Squares


altayhunoglu:
https://leetcode.com/problems/word-squares/

Varsayılan code:

```
class Solution {
    public List<List<String>> wordSquares(String[] words) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380526-176ea000-b5eb-11eb-89e2-bab1dcafd162.png)
![image](https://user-images.githubusercontent.com/60903744/118380529-1dfd1780-b5eb-11eb-94cb-1dadf748d93e.png)
![image](https://user-images.githubusercontent.com/60903744/118380532-23f2f880-b5eb-11eb-80a7-be17cf61f50a.png)

PR #375, opened 2021-07-18
425. Word Squares

altayhunoglu:
Resolves: https://github.com/spiralgo/algorithms/issues/220

425_valid_word_square

A word square matrix is naturally a symmetric matrix.
It means when we try the `candidate` words starting from the top row to the bottom row, the symmetric order will occur eventually, like in the above picture.
Therefore, we can interpret this situation as a `constraint`, and a `constraint` is nothing but a guide to pruning a solution tree. 
As this question asks for `all possible solutions` to construct a word square and there occurs a `constraint` that helps us prune the tree, a kind of `backtracking technique` inevitably comes to mind.

In each insertion of the candidate words into the matrix, a `prefix` (which is the constraint at the same time) is built naturally. 
Like in the picture:
-  When we insert `ball` into the matrix as the first candidate, we are obliged to insert a word starting with the _prefix_ `a` into the second row. 
- After we insert the only suitable word "area", we are obliged to insert a word starting with the _prefix_ `le` into the third row.
 
We could use a HashMap kind of data structure to store and retrieve the list of the words starting with the prefix needed.
We would use the prefix as a key, and the list of the words as the value.
 
 But when we mention `prefix` and `re(trie)ve`, a Trie (prefix tree) data structure might be the perfect match.
 
 The solution in the code is the fastest one on LeetCode. I just modularized and cleaned it to make it more readable.
 Although the code is quite brilliant, there are some parts that disturb me. 
 
  For example, we need to fill an array with the same prefix tree to prune the root after each step. Sure, it helps to fasten the search on the tree, but there might be a leaner way about memory. 
  
  But for now, the solution is quite fine, according to me. When we turn back for this question to refactor it, we can think about making it leaner.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/WordSquaresTrieBacktracking.java
+++ b/src/main/java/algorithms/curated170/hard/WordSquaresTrieBacktracking.java
@@ -1,0 +1,93 @@
+package algorithms.curated170.hard;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class WordSquaresTrieBacktracking {
+   public static class Solution {
+       public final static int LETTER_a = 97;
+
+       public List<List<String>> wordSquares(String[] words) {
+           TrieNode root = buiildTrie(words);
+           int n = words[0].length();
+           TrieNode[] trieNodeState = new TrieNode[n];
+           Arrays.fill(trieNodeState, root);
+           List<List<String>> ans = new ArrayList<>();
+           dfs(0, new char[n][n], ans, trieNodeState);
+           return ans;
+       }
+       void dfs(int rowIndex, char[][] matrix, List<List<String>> ans, TrieNode[] trieNodeState) {
+           if (rowIndex == matrix.length) {
+               ans.add(convertMatrixToList(matrix, rowIndex));
+               return;
+           }
+           TrieNode[] trieNodeStateNext = new TrieNode[trieNodeState.length];
+           for (String candidate : trieNodeState[rowIndex].list) {
+
+               if (isSymmetric(matrix, candidate, rowIndex, trieNodeState, trieNodeStateNext)){
+                   dfs(rowIndex + 1, matrix, ans, trieNodeStateNext);
+               }
+           }
+       }
+
+       private boolean isSymmetric(char[][] matrix, String candidate, int rowIndex,
+                                   TrieNode[] trieNodeState, TrieNode[] trieNodeStateNext) {
+
+           int i = rowIndex;
+           while (i < matrix.length) {
+               char ch = candidate.charAt(i);
+               matrix[rowIndex][i] = ch;
+               if (i > rowIndex) {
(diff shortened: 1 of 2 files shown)


src/main/java/algorithms/curated170/hard/WordSquaresTrieBacktracking.java
```diff
@@ -0,0 +1,93 @@
...
+
+           int i = rowIndex;
+           while (i < matrix.length) {
+               char ch = candidate.charAt(i);
+               matrix[rowIndex][i] = ch;
+               if (i > rowIndex) {
+                   matrix[i][rowIndex] = ch;
+                   int letterIndex = letterIndex(ch);
+                   if (trieNodeState[i].children[letterIndex] == null) break;
+                   trieNodeStateNext[i] = trieNodeState[i].children[letterIndex];
```
ErdemT09, line 45: It seems that `trieNodeStateNext` doesn't do anything. If I'm not mistaken, we can remove it.

src/main/java/algorithms/curated170/hard/WordSquaresTrieBacktracking.java
```diff
@@ -0,0 +1,93 @@
...
+       }
+       void dfs(int rowIndex, char[][] matrix, List<List<String>> ans, TrieNode[] trieNodeState) {
+           if (rowIndex == matrix.length) {
+               ans.add(convertMatrixToList(matrix, rowIndex));
+               return;
+           }
+           TrieNode[] trieNodeStateNext = new TrieNode[trieNodeState.length];
+           for (String candidate : trieNodeState[rowIndex].list) {
+
+               if (isSymmetric(matrix, candidate, rowIndex, trieNodeState, trieNodeStateNext)){
```
ErdemT09, line 28: Note: `isSymmetric()` method handles the backtracking part too. As we call the method, it overrides the character matrix with candidate string. If it's okay, we continue with it. If it's not symmetric, we don't. Either way, it get's overriden in the next function call.

src/main/java/algorithms/curated170/hard/WordSquaresTrieBacktracking.java
altayhunoglu: It has a function as a kind of "backtracking" to the older "trieNodeState" keeping the older immutable.
But, you can try removing of course.

ErdemT09: I didn't see it being inputted into dfs. It makes more sense now, as we are traversing after some prefix already gone through.

ErdemT09, line 45: This allows duplicate usage of words, as the problem allows.

ErdemT09 (approved):
Such string algorithms are something that we have not worked on in this repository. This is a good example for further algorithms we might examine such suffix arrays and trees.

PR #376, opened 2021-07-19
787. Cheapest Flights within K Stops

ErdemT09:
Problem: https://leetcode.com/problems/cheapest-flights-within-k-stops/

We can interpret this problem as a graph problem where we have to find the shortest path, set of "flights" from the source city to the destination. The graph here is a directed, weighted graph. So, it's only the most intuitive to use shortest path algorithms here. However, the number of edges we can traverse is limited by the number `K`, We can't visit more than `K+1` edges and `K+2` vertices 

### .Approach-1:  BFS:
I think this is the most simple one. In order not to visit more than limited number of nodes, we simply keep a check of how many layers of BFS we have done, as one can perhaps recall from #180. We do no more than K+1 layers of BFS, and eventually check for the minimum distance for the destination, or return -1 i we can't find it within K+1 layers of BFS.

### Approach 2: DFS:
This is also a pretty simple modification. Only making K stops is also equivalent to applying the DFS algorithm to a depth of K+1. The minimum distance is then to be found within these searches. Since DFS contains making the same searches again, we can memoize these operations. Besides, we can prune our decision tree. If the current cost we get from DFS is more than the minimum global cost that we have found, we can avoid going any deeper.

### Approach 3: Belmann-Ford Algorithm:
Belmann-Ford algorithm works by "relaxing" the edges. "Relaxation" simply means that if a distance to node would be updated iff the distance to a previous node + the edge weight yields a less distance. Normally, this relaxation process is done `|V|-1` times, which is the maximum amount of nodes to be traversed in order to get to the destination. We can decrease this value to K+1 in order to only relax edges this many times, thus making at most K stops.

In implementation, In order not to update the current distance array, we can create a temporary array. However, already having created 2 arrays and switching between these two for even and odd iterations. Doing this with bit-gates is pretty lean. It's not legible at first sight, but it's pretty great to use once learned.

### Approach 4: Dijkstra's Algorithm:
The problem with the usage of Dijkstra's algorithm here is that, when we pick the shortest distance for a node and the path doesn't satisfy having K stops, there is no way of backtracking and picking the shortest path less than K stops. For this, we not only consider the distances of nodes but also the number of stops they have, thus storing multiple paths. When we are adding a node to the priority queue, we only add if it satisfies having a shorter distance  or else having less number of stops. Either way, we update the current number of stops to the node, so that we prune the decision tree for adding the same node after adding this one.

### DIFF
--- a/src/main/java/algorithms/medium/CheapestFlightsWithinKStops.java
+++ b/src/main/java/algorithms/medium/CheapestFlightsWithinKStops.java
@@ -1,0 +1,91 @@
+package algorithms.medium;
+
+import java.util.PriorityQueue;
+
+public class CheapestFlightsWithinKStops {
+
+    public int findCheapestPrice(int n, int[][] flights, int src, int dst, int k) {
+        int[][][] graph = createGraph(flights, n);
+        return dijkstra(graph, src, dst, k, n);
+    }
+
+    int[][][] createGraph(int[][] arcs, int n){
+        int[][][] graph = new int[n][][];
+        
+        int[] outDegree = new int[n + 1];
+        for (int[] arc : arcs) {
+            outDegree[arc[0]]++;
+        }
+
+        for (int i = 0; i < n; i++) {
+            graph[i] = new int[outDegree[i]][2];
+        }
+        int[] currIdx = new int[n];
+        for (int[] arc : arcs) {
+            graph[arc[0]][currIdx[arc[0]]][0] = arc[1];
+            graph[arc[0]][currIdx[arc[0]]++][1] = arc[2];
+        }
+        return graph;
+    }
+
+    int dijkstra(int[][][] graph, int src, int dst, int k, int n)
+    {
+        int[] dist = new int[n];
+        int[] steps = new int[n];
+        PriorityQueue<int[]> pq = new PriorityQueue<>((a,b) -> a[1]-b[1]);
+        for(int i = 0; i<n; i++)
+        {
+            dist[i] = Integer.MAX_VALUE;
+            steps[i] = Integer.MAX_VALUE;
+        }
+        dist[src]= 0;
+        steps[src] = 0;
+        pq.offer(new int[]{src, 0, 0});
+
+        
+        while(!pq.isEmpty())
+        {
+            int[] info = pq.poll();
+            int vertex = info[0];
+            int cost = info[1];
+            int stops = info[2];
+            
+            if(vertex == dst)
+            {
+                return cost;
+            }
+            if(stops == k+1)
+            {
+                continue;
+            }
+
+            for(int[] neighbor : graph[vertex])
+            {
+                int next = neighbor[0];
+                int alt = cost + neighbor[1];
+                if(alt < dist[next])
+                {
+                    pq.offer(new int[]{next, alt, stops+1});
+                    dist[next]  = alt;
+                }
+                else if(stops < steps[next]){
+                    pq.offer(new int[]{next, alt, stops+1});
+                }
+                steps[next] = stops;
+            }
+        }
+        return -1;
+    }
+    public static void main(String[] args) {
+        int[][] flights = {
+            {0,1,200},
+            {0,2,500},
+            {1,2,250}
+        };
+        var solution = new CheapestFlightsWithinKStops();
+        System.out.println(solution.findCheapestPrice(3, flights, 0, 2, 0));
+
+        flights = new int[][]{{0,1,1},{0,2,5},{1,2,1},{2,3,1}};
+        System.out.println(solution.findCheapestPrice(4, flights, 0, 3, 1));
+    }
+}
--- a/src/main/java/algorithms/medium/CheapestFlightsWithinKStops.java
+++ b/src/main/java/algorithms/medium/CheapestFlightsWithinKStops.java
@@ -4,14 +4,14 @@ import java.util.PriorityQueue;
 
 public class CheapestFlightsWithinKStops {
 
-    public int findCheapestPrice(int n, int[][] flights, int src, int dst, int k) {
+    public int findCheapestPrice(int n, int[][] flights, int src, int dst, final int K) {
         int[][][] graph = createGraph(flights, n);
-        return dijkstra(graph, src, dst, k, n);
+        return dijkstra(graph, src, dst, K, n);
     }
 
-    int[][][] createGraph(int[][] arcs, int n){
+    int[][][] createGraph(int[][] arcs, int n) {
         int[][][] graph = new int[n][][];
-        
+
         int[] outDegree = new int[n + 1];
         for (int[] arc : arcs) {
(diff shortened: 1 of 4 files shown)


src/main/java/algorithms/medium/CheapestFlightsWithinKStops.java
```diff
@@ -0,0 +1,84 @@
...
+                int nextVertex = arc[0];
+                int altDist = cost + arc[1];
+
+                if (altDist < dist[nextVertex]) {
+                    pq.offer(new int[] { nextVertex, altDist, stops + 1 });
+                    dist[nextVertex] = altDist;
+                } else if (stops < currStops[nextVertex]) {
+                    pq.offer(new int[] { nextVertex, altDist, stops + 1 });
+                }
+                currStops[nextVertex] = stops;
```
ErdemT09, line 60: I wasn't able to rigorously prove why this is efficient. It might actually even be a heuristic. I will let it stay for now.

altayhunoglu: As far as I understand; it gives a new consideration chance to an already visited node; if the node has a lesser stop value than it had before. 

 The reason is that the total distance to the node from the source might have been changed with the new configuration of the state.

  (By the way, It adds the same node multiple times into the PQ rather than updating the existing values of the node. Because otherwise, we would pay an extra search cost to find the node in the PQ.)

ErdemT09: Then let's think of some ways the next vertex can be handled, when it's going to be traversed to from some neighbor nodes:
```
1. Path: distance = 9, stops = 6
2. Path: distance = 13, stops 4
3. Path: distance = 11, stops 5
4. Path: distance = 20, stops 4
```

Of these 4 paths, only the 3. would not be added. But what if the destination was after this node and k equaled to 7? Then the most efficient path would have been the third one, which we did not add to the priority queue.

altayhunoglu: I think we need to think in terms of nodes and edges instead of the distances in the first place.
Could you draw the graph according to these distances and stops you mentioned, please?

ErdemT09: I meant something like this.

altayhunoglu: Thank you, Erdem. I will have a loot at it now.

ErdemT09: Actually, the node with length 9 will be polled out before the previous path to it with length 10. But I can find a new example that doesn't have such a property.

ErdemT09: As in here.

altayhunoglu: Then our test-case would be this, right?
```js
17
[[0,1,1],[0,2,1],[0,3,2],[0,4,3], [1,5,2], [5,6,1],[6,7,2], [7,8,1], [8,11,12], [2,9,4],[9,10,4],[10,11,14],[3,12,1],[12,13,3],[13,14,4], [14,11,11], [4,15,6], [15,16,7], [16,11,11]]
0
11
1
```
yes

ErdemT09: > Then our test-case would be this, right?
> ```js
> 17
> [[0,1,1],[0,2,1],[0,3,2],[0,4,3], [1,5,2], [5,6,1],[6,7,2], [7,8,1], [8,11,12], [2,9,4],[9,10,4],[10,11,14],[3,12,1],[12,13,3],[13,14,4], [14,11,11], [4,15,6], [15,16,7], [16,11,11]]
> 0
> 11
> 1
> ```
> yes
> 

Destination is the node below it, with an arc to it of weight 100.

altayhunoglu: Ah, oke, I fix it like this. Could you check, please?
```js
18
[[0,1,1],[0,2,1],[0,3,2],[0,4,3], [1,5,2], [5,6,1],[6,7,2], [7,8,1], [8,11,12], [2,9,4],[9,10,4],[10,11,14],[3,12,1],[12,13,3],[13,14,4], [14,11,11], [4,15,6], [15,16,7], [16,11,11], [11,12,100]]
0
12
7
```

ErdemT09:
```
18
[[0,1,1],[0,2,1],[0,3,2],[0,4,3], [1,5,2], [5,6,1],[6,7,2], [7,8,1], [8,11,12], [2,9,4],[9,10,4],[10,11,14],[3,12,1],[12,13,3],[13,14,4], [14,11,11], [4,15,6], [15,16,7], [16,11,11], [11,12,100]]
0
12
6
```
It should have 6 stops. When I run this case, LeetCode outputs 3 though.

altayhunoglu: I think this is the correct case:
```java
18
[[0,1,1],[0,2,1],[0,3,2],[0,4,6],[1,5,2], [5,6,1],[6,7,2], [7,8,1], [8,11,12],[2,9,4],[9,10,4],[10,11,14],[3,12,1],[12,13,3],[13,14,4], [14,11,11], [4,15,6], [15,16,7], [16,11,11], [11,12,100]]
0
12
7
```

But, then again the output is 3.

ErdemT09: For
```
18
[[0,1,1],[0,2,1],[0,3,2],[0,4,3],[1,5,2],[5,6,1],[6,7,2],[7,8,1],[8,11,12],[2,9,4],[9,10,4],[10,11,14],[3,12,1],[12,13,3],[13,14,4],[14,11,11],[4,15,6],[15,16,7],[16,11,11],[11,17,100]]
0
17
5
```
It gave the wrong answer. I think we can conclude at this point that this line of code is just an imperfect heuristic, that somehow works through all the test cases of LeetCode. Dijkstra wasn't appropriate for this problem anyway.

altayhunoglu: Oke, then I will just approve. The others look fine. When we turn back, we can have a look again.

altayhunoglu (approved):
Thank you for your valuable collaboration Erdem.

altayhunoglu (approved):
Thank you for your valuable collaboration Erdem.

ErdemT09:
> Thank you for your valuable collaboration Erdem.

Thank you for you collaboration too. I think we shouldn't try showing contradictions in algorithms by just brute counter-testing. That takes too much effort sometimes. Proving by contradictions in less direct manners is usually better in terms of logic too.

Issue #216, opened 2021-05-15
1259. Handshakes That Don't Cross


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int numberOfWays(int num_people) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380356-d0cc7600-b5e9-11eb-8953-1aa873a6db56.png)
![image](https://user-images.githubusercontent.com/60903744/118380362-d75aed80-b5e9-11eb-8236-124cc777ad9a.png)
![5125_example_3](https://user-images.githubusercontent.com/60903744/118380375-ea6dbd80-b5e9-11eb-84de-e2b33b1f7b1c.png)
![image](https://user-images.githubusercontent.com/60903744/118380377-f48fbc00-b5e9-11eb-9fde-2bf25eee885c.png)

altayhunoglu:
https://www.youtube.com/watch?v=0pTN0qzpt-Y&ab_channel=KeertiPurswaniKeertiPurswani

ErdemT09:
In each handshake, we divide the "hands" into two different groups, and calculate how many such division we can make in turn in those divisions. Like in #338, this is related to Catalan numbers somehow, I think.

ErdemT09:
The recursive formula should look something like this:
handshakes(n) := hs(n)
hs(n) := hs(0)*hs(n-2) + hs(2)*hs(n-4) + hs(4)*hs(n-6) ... + hs(n-2)*hs(0)

PR #378, opened 2021-07-20
1259. Handshakes That Don't Cross

ErdemT09:
Resolves: #216 

## Algorithm:
What happens here is that each hand shake divides the group into 2 distinct groups. The total number of handshakes for this division would then be handshakes for the one group times the handshakes for the other group, by the [multiplication principle](https://en.wikipedia.org/wiki/Rule_of_product).  To count the number of such divisions, we can choose a person like `person 1` to be the pivot. This person can shake hands with person `2, 4, 6, ..., n-2, n`. Shaking hands with person k would seperate the group into the groups of size k-2 and n-k. This also equivalent to calling the handshake function with values `0, 2, 4, ..., n-4, n-2`. We can derive a recurrence relation from this:

The recursive formula should look something like this:
```
handshakes(n) := hs(n)
hs(0) := 1
hs(2) := 1
hs(n) := sum( hs(2*a)*hs(n-2*(a+1)), 0≤a≤(n/2-1) ) 
= hs(0)*hs(n-2) + hs(2)*hs(n-4) + hs(4)*hs(n-6) ... + hs(n-2)*hs(0)
From this, we can check:
hs(4) = hs(0)*hs(2)+hs(2)*hs(0) = 2 (true)
hs(6) = hs(0)*hs(4)+hs(2)*hs(2)+hs(4)*hs(0) = 5 (true)
where hs(n) ≡ Catalan(n/2)
```
We can implement this using recursion with memoization as well as dynamic programming.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/HandshakesThatDontCross.java
+++ b/src/main/java/algorithms/curated170/hard/HandshakesThatDontCross.java
@@ -1,0 +1,28 @@
+package algorithms.curated170.hard;
+
+public class HandshakesThatDontCross {
+
+    private static long[] cache = new long[501];
+    static {
+        cache[0] = cache[1] = 1;
+    }
+
+    private static final int MOD = (int) 1e9 + 7;
+
+    public int numberOfWays(int n) {
+        return (int) catalan(n / 2);
+    }
+
+    private long catalan(int n) {
+        if (cache[n] != 0) {
+            return cache[n];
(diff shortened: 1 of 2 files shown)


altayhunoglu (approved):
Thank you, Erdem. A wonderful application of Catalan Numbers.

altayhunoglu:
IMG-1713

Issue #214, opened 2021-05-15
527. Word Abbreviation


altayhunoglu:
https://leetcode.com/problems/word-abbreviation/

Varsayılan kod:

```
class Solution {
    public List<String> wordsAbbreviation(List<String> words) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380293-48e66c00-b5e9-11eb-893f-b6a0c0441ba5.png)
![image](https://user-images.githubusercontent.com/60903744/118380300-500d7a00-b5e9-11eb-9285-b9bbc01486ed.png)

PR #379, opened 2021-07-21
527. Word Abbreviation

ErdemT09:
Resolves: #214 

## Algorithm: 
In order to fix the duplicate abbreviation situation stated in the problem, we simply begin by abbreviating all the words as they would be without considering the other words. For each abbreviation, we put it as a key into a map and add the word that it is abbreviated from. Thus, this map stores which abbreviations correspond to which words, so we can see the duplicates.

After that, for each word, we check if the entry of its abbreviation in the map contains duplicate (its list having a size >1). If so, we distribute these words into new abbreviations by incrementing their prefix. For the words `"spiralgo"`, the old abbreviation "s6o" will become "sp5o". If a word like "smnbgfdo" had the same abbreviation "s6o", we would also remake it to "sm5o". After doing this re-abbreviating, we check the word again. If there are still duplicates (for example "spwwwwwo" becoming "sp5o" too) we reabbreviate again until there are no duplicates left, or we run out of letters.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/WordAbbreviation.java
+++ b/src/main/java/algorithms/curated170/hard/WordAbbreviation.java
@@ -1,0 +1,67 @@
+package algorithms.curated170.hard;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+
+public class WordAbbreviation {
+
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Very nice explanation using the word "Spiralgo" and a well-modularized code.
Nothing further to add on my side.
Thanks again Erdem.

Issue #215, opened 2021-05-15
1548. The Most Similar Path in a Graph


altayhunoglu:
https://leetcode.com/problems/the-most-similar-path-in-a-graph/

Varsayılan kod:

```
class Solution {
    public List<Integer> mostSimilar(int n, int[][] roads, String[] names, String[] targetPath) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380315-74695680-b5e9-11eb-9928-b90ed5ac1d3a.png)
![image](https://user-images.githubusercontent.com/60903744/118380321-81864580-b5e9-11eb-9357-8e9fe5af032e.png)
![image](https://user-images.githubusercontent.com/60903744/118380326-8ea33480-b5e9-11eb-83ec-0ecd30df51ca.png)
![image](https://user-images.githubusercontent.com/60903744/118380329-94991580-b5e9-11eb-9cca-51441778fb31.png)
![image](https://user-images.githubusercontent.com/60903744/118380331-9a8ef680-b5e9-11eb-8287-4923539208a4.png)
![image](https://user-images.githubusercontent.com/60903744/118380334-a1b60480-b5e9-11eb-8153-3c57ccde1183.png)
![image](https://user-images.githubusercontent.com/60903744/118380336-a7abe580-b5e9-11eb-90c2-e83a12264407.png)

PR #380, opened 2021-07-21
1548. The Most Similar Path in a Graph

ErdemT09:
Resolves: #215 

## Approach 1-Dynamic Programming:
We should notice the recurrence relation:
The most similar path of length `T` is the concatenation of some city `c==targetPath[0]` and the most similar path that can start from that city, of length `T-1`.
Written in dynamic programming form:
`dp[t][c] = min{dp[t+1][next]} + targetPath[t] == c ? 0 : 1`, where `t` is the index of the target path we are considering and `next`∈ All cities connected to `c`.
We can first calculate `dp[lastIndex][c]` for all cities. `dp[lastIndex-1][c]` would then be equal to minimum of `dp[lastIndex][next]` for all cities connected `c` (of course, added with 1 or 0 for the edit distance). We can thus loop our way back to the 0th index. 

After that, we create the path simply by retracing the dp array with the minimum values and the previous city we visited.

## Approach 2-Priority Queue Implemented with Double-ended Queue:
Here, it is naturally intuitive to place all the paths in a priority queue and keep poll out the one with the least edit distance.  However, since we are allowed to return back to the same node after visiting it, this approach would render inefficient with a possible worst case complexity of `O(p·log p)`, `p=number of paths`. We won't get the benefits that an algorithm like Dijkstra's would have, as we can't have anything like distance array. So, the most convenient option is to use a dequeue as a sort of a priority queue. Each time, if some node on the path is equal to its corresponding name in the `targetPath`, we put its new path on the start of the queue, else we put it to the end of the queue. We remove paths from the queue by removing the first element, so the relatively best path. At the end, we would poll out the path with the length of the `targetPath` that is the most similar. I won't try to prove this, it's however pretty intuitive.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/TheMostSimilarPathInAGraph.java
+++ b/src/main/java/algorithms/curated170/hard/TheMostSimilarPathInAGraph.java
@@ -12,24 +12,6 @@ public class TheMostSimilarPathInAGraph {
     int targetLength;
     int[][] distancesAtCities;
 
-    private void calculateDistancesBackwards() {
-        String targetCity;
-        for (int t = targetLength - 2; t >= 0; t--) {
-            targetCity = targetPath[t];
-            for (int c = 0; c < n; c++) {
-                String curCity = names[c];
-                if (!curCity.equals(targetCity))
-                    distancesAtCities[t][c] = 1;
-                int minNextValue = Integer.MAX_VALUE;
-                for (int nextCity = 0; nextCity < n; nextCity++) {
-                    if (isAdjacent[nextCity][c] == true)
-                        minNextValue = Math.min(distancesAtCities[t + 1][nextCity], minNextValue);
-                }
-                distancesAtCities[t][c] += minNextValue;
-            }
-        }
-    }
-
     public List<Integer> mostSimilar(int n, int[][] roads, String[] names, String[] targetPath) {
 
         this.n = n;
@@ -51,6 +33,41 @@ public class TheMostSimilarPathInAGraph {
         calculateDistancesBackwards();
 
         return evaluateMostSimilarPath();
+    }
+
+    private void setLastDistances() {
+        String targetCity = targetPath[targetLength - 1];
+        for (int i = 0; i < n; i++) {
+            String curCity = names[i];
+            if (!targetCity.equals(curCity)) {
+                distancesAtCities[targetLength - 1][i] = 1;
+            }
+        }
+    }
+
+    private void calculateDistancesBackwards() {
+        String targetCity;
+
+        for (int t = targetLength - 2; t >= 0; t--) {
+            targetCity = targetPath[t];
+
+            for (int c = 0; c < n; c++) {
+                String curCity = names[c];
+
+                if (!curCity.equals(targetCity)) {
+                    distancesAtCities[t][c] = 1;
+                }
+
+                int minNextValue = Integer.MAX_VALUE;
+                for (int nextCity = 0; nextCity < n; nextCity++) {
+                    if (isAdjacent[nextCity][c] == true) {
+                        minNextValue = Math.min(distancesAtCities[t + 1][nextCity], minNextValue);
+                    }
+                }
+
+                distancesAtCities[t][c] += minNextValue;
+            }
+        }
     }
 
     private List<Integer> evaluateMostSimilarPath() {
@@ -92,15 +109,4 @@ public class TheMostSimilarPathInAGraph {
         mostSimilarPath.add(curMinCity);
         addNextVertices(mostSimilarPath, curMinCity, idx + 1);
     }
-
-    private void setLastDistances() {
-        String targetCity = targetPath[targetLength - 1];
-        for (int i = 0; i < n; i++) {
-            String curCity = names[i];
-            if (!targetCity.equals(curCity)) {
-                distancesAtCities[targetLength - 1][i] = 1;
-            }
-        }
-    }
-
 }
--- a/src/main/java/algorithms/curated170/hard/TheMostSimilarPathInAGraph.java
+++ b/src/main/java/algorithms/curated170/hard/TheMostSimilarPathInAGraph.java
@@ -1,0 +1,92 @@
+package algorithms.curated170.hard;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TheMostSimilarPathInAGraph {
+
+    public List<Integer> mostSimilar(int n, int[][] roads, String[] names, String[] targetPath) {
+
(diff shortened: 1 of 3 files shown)


altayhunoglu:
Both solutions are quite brilliant Erdem.

I strongly recommend introducing a City class to make it easy to understand and clear.
I refactored the code a little bit, but it would be better if you look over it.

```
class Solution {
     
  public List<Integer> mostSimilar(int n, int[][] roads, String[] names, String[] targetPath) {
        int t = targetPath.length;
        if (t == 0) {
            return new ArrayList<>();
        }

        int[][] graph = createGraph(n, roads);

        Deque<City> paths = new ArrayDeque<>();

        setPathStarts(n, names, targetPath, paths);

        return buildPathWithQueue(graph, paths, names, targetPath, n);
    }

    private List<Integer> buildPathWithQueue(int[][] graph, Deque<City> paths, String[] names, String[] targetPath,
            int n) {
        int t = targetPath.length;

        boolean[][] seen = new boolean[n][t];
        while (!paths.isEmpty()) {
            City city = paths.poll(); 
            List<Integer> currPathNodes = city.currPathNodes;

            int length = currPathNodes.size(); 
            
            if (length == t) {
                return currPathNodes;
            }

            List<Integer> nextPathNodes;
            for (int neighbor : graph[city.cityIndex]) {
                if (seen[neighbor][length] == true) {
                    continue;
                }
                seen[neighbor][length] = true;

                nextPathNodes = new ArrayList<>(currPathNodes);
                nextPathNodes.add(neighbor);

                City nextPath = new City(neighbor, nextPathNodes);

                if (names[neighbor].equals(targetPath[length])) {
                    paths.addFirst(nextPath);
                } else {
                    paths.addLast(nextPath);
                }
            }
        }
        return new ArrayList<>();
    }

    private void setPathStarts(int n, String[] names, String[] targetPath, Deque<City> q) {
        for (int i = 0; i < n; i++) {
            List<Integer> list = new ArrayList<>();
            list.add(i);
            City initialCity = new City(i, list);
            if (targetPath[0].equals(names[i])) {
                q.addFirst(initialCity);
            } else {
                q.addLast(initialCity);
            }
        }
    }

    private int[][] createGraph(int n, int[][] roads) {
        int[][] graph = new int[n][];

        int[] adjacentCount = new int[n];
        for (int[] road : roads) {
            adjacentCount[road[0]]++;
            adjacentCount[road[1]]++;
        }

        for (int i = 0; i < n; i++) {
            graph[i] = new int[adjacentCount[i]];
        }
        int[] currIdx = new int[n];
        for (int[] road : roads) {
            graph[road[0]][currIdx[road[0]]++] = road[1];
            graph[road[1]][currIdx[road[1]]++] = road[0];
        }
        return graph;
    }
     private class City{
        City(int cityIndex, List<Integer> currPathNodes){
            this.cityIndex = cityIndex;
            this.currPathNodes = currPathNodes;
        }
        int cityIndex;
        List<Integer> currPathNodes; 
    }
}

```

ErdemT09:
Changes added.

altayhunoglu:
We can apply your `graph` idea for the DP solution in order to prevent traversing all the elements in the "name" array to ask if they are adjacent.
  It would fasten the code remarkably.
  Graph array can be considered as an adjacency matrix.
 Therefore, we can refactor like this:

 ```java
class Solution {
     
    int n;
    String[] names;
    String[] targetPath;
    int targetLength;
    int[][] distancesAtCities;
    int[][] graph;

    public List<Integer> mostSimilar(int n, int[][] roads, String[] names, String[] targetPath) {

        this.n = n;
        this.names = names;
        this.targetPath = targetPath;
        graph = createGraph(n, roads);
       

        targetLength = targetPath.length;
        distancesAtCities = new int[targetLength][n];

        setLastDistances();

        calculateDistancesBackwards();

        return evaluateMostSimilarPath();
    }
private int[][] createGraph(int n, int[][] roads) {
        int[][] graph = new int[n][];

        int[] adjacentCount = new int[n];
        for (int[] road : roads) {
            adjacentCount[road[0]]++;
            adjacentCount[road[1]]++;
        }

        for (int i = 0; i < n; i++) {
            graph[i] = new int[adjacentCount[i]];
        }
        int[] currIdx = new int[n];
        for (int[] road : roads) {
            graph[road[0]][currIdx[road[0]]++] = road[1];
            graph[road[1]][currIdx[road[1]]++] = road[0];
        }
        return graph;
    }
    private void setLastDistances() {
        String targetCity = targetPath[targetLength - 1];
        for (int i = 0; i < n; i++) {
            String curCity = names[i];
            if (!targetCity.equals(curCity)) {
                distancesAtCities[targetLength - 1][i] = 1;
            }
        }
    }

    private void calculateDistancesBackwards() {
        String targetCity;

        for (int t = targetLength - 2; t >= 0; t--) {
            targetCity = targetPath[t];

            for (int c = 0; c < n; c++) {
                String curCity = names[c];

                if (!curCity.equals(targetCity)) {
                    distancesAtCities[t][c] = 1;
                }
             
                int minNextValue = Integer.MAX_VALUE;
             
                for (int nextCity : graph[c]) {
                    
                        minNextValue = Math.min(distancesAtCities[t + 1][nextCity], minNextValue);
                    
                }

                distancesAtCities[t][c] += minNextValue;
            }
        }
    }

    private List<Integer> evaluateMostSimilarPath() {
        List<Integer> mostSimilarPath = new ArrayList<>(targetLength);
        int prevCity = 0;

        for (int i = 1; i < n; i++) {
            if (distancesAtCities[0][i] < distancesAtCities[0][prevCity]) {
                prevCity = i;
            }
        }
        mostSimilarPath.add(prevCity);

        addNextVertices(mostSimilarPath, prevCity, 1);

        return mostSimilarPath;
    }

    private void addNextVertices(List<Integer> mostSimilarPath, int prevCity, int idx) {

        if (idx == targetLength) {
            return;
        }

        int curMinCity = -1;
        int curMinDist = Integer.MAX_VALUE;

          for (int nextCity : graph[prevCity]) {

            if (distancesAtCities[idx][nextCity] < curMinDist) {
                curMinDist = distancesAtCities[idx][nextCity];
                curMinCity = nextCity;
            }
        }

        mostSimilarPath.add(curMinCity);
        addNextVertices(mostSimilarPath, curMinCity, idx + 1);
    }
} 
```

ErdemT09:
> It would fasten the code remarkably.
I thought adjacency matrix was okay, since it was only booleans.

altayhunoglu (approved):
I think this question was really hard.
 But, you nailed it meticulously.
 It is a great pleasure to work with you.

Issue #240, opened 2021-05-15
644. Maximum Average Subarray II


altayhunoglu:
https://leetcode.com/problems/maximum-average-subarray-ii/

Varsayılan kod:

```
class Solution {
    public double findMaxAverage(int[] nums, int k) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380942-a204ce80-b5ee-11eb-841e-c562b982c284.png)
![image](https://user-images.githubusercontent.com/60903744/118380945-a7faaf80-b5ee-11eb-954f-b5ea993bf65c.png)
![image](https://user-images.githubusercontent.com/60903744/118380948-ac26cd00-b5ee-11eb-8c3b-6f462b6bbf67.png)

PR #381, opened 2021-07-23
644. Maximum Average Subarray II

ErdemT09:
Resolves: #240 

## Algorithm:
The problem here is pretty similar to other binary search problems we had like #369 and #368. We pick some target value, try to find if there is some subarray whose length ≥ k that at least has the target value as average. If it does, we can look for higher values. Else, we need to look for lower values. 

While looking if there is such a subarray, we look for differences between averages. If `s_i` is equal to sum of all elemenets up to some index `i`, a subarray with average `v` up to that would have the sum `i·v`. `s_i-i·v - s_j-j·v ≥ 0` would prove that there is a subarray of length `i-j` that has an average ≥ `v`. Since the first term also includes the second term in its total, we have to subtract it to get the difference from the average subarray. Finding some `j, i-j ≥ k` would imply that such a subarray exists.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/MaximumAverageSubarrayII.java
+++ b/src/main/java/algorithms/curated170/hard/MaximumAverageSubarrayII.java
@@ -22,15 +22,23 @@ public class MaximumAverageSubarrayII {
         return lo;
     }
     
-    public boolean canFind(int[] sums, int k, double v) {
+          public boolean canFind(int[] sums, int k, double mid) {
         double sum = 0;
-        double minPreSum = 0;
+        double preSum=0, minPreSum = 0;
         for (int i = k-1; i < sums.length; i++) {
-            sum = sums[i] - (i+1) * v;
-            if (i-k >= 0) minPreSum = Math.min(minPreSum, sums[i-k] - (i-k+1) * v);
-            if (sum - minPreSum > 0) return true;
+            sum = getSumInRange(i,0, sums, mid);
+           
+            if (i-k >= 0){
+                 preSum = getSumInRange(i,k, sums, mid);
+                 minPreSum = Math.min(minPreSum, preSum);
+            } 
+            if (sum - minPreSum >= 0) return true;
         }
         return false;
     }
+    
+     double getSumInRange(int i, int k, int[] sums, double mid){
+        return sums[i-k] - (i-k+1) * mid;
+     }
 
 }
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/hard/MaximumAverageSubarrayII.java
```diff
@@ -0,0 +1,36 @@
...
+        return lo;
+    }
+    
+    public boolean canFind(int[] sums, int k, double v) {
+        double sum = 0;
+        double minPreSum = 0;
+        for (int i = k-1; i < sums.length; i++) {
+            sum = sums[i] - (i+1) * v;
+            if (i-k >= 0) minPreSum = Math.min(minPreSum, sums[i-k] - (i-k+1) * v);
+            if (sum > minPreSum) return true;
```
ErdemT09, line 31:
```suggestion
            if (sum - minPreSum > 0) return true;
```

src/main/java/algorithms/curated170/hard/MaximumAverageSubarrayII.java
```diff
@@ -0,0 +1,36 @@
...
+    public boolean canFind(int[] sums, int k, double v) {
+        double sum = 0;
+        double minPreSum = 0;
+        for (int i = k-1; i < sums.length; i++) {
+            sum = sums[i] - (i+1) * v;
+            if (i-k >= 0) minPreSum = Math.min(minPreSum, sums[i-k] - (i-k+1) * v);
+            if (sum - minPreSum > 0) return true;
+        }
+        return false;
+    }
```
altayhunoglu, line 34: It was a little bit difficult for me to understand this question.
Adding some extra modularity helped me a lot.
So I may suggest that.


```suggestion
          public boolean canFind(int[] sums, int k, double mid) {
        double sum = 0;
        double preSum=0, minPreSum = 0;
        for (int i = k-1; i < sums.length; i++) {
            sum = getSumInRange(i,0, sums, mid);
           
            if (i-k >= 0){
                 preSum = getSumInRange(i,k, sums, mid);
                 minPreSum = Math.min(minPreSum, preSum);
            } 
            if (sum - minPreSum >= 0) return true;
        }
        return false;
    }
    
     double getSumInRange(int i, int k, int[] sums, double mid){
        return sums[i-k] - (i-k+1) * mid;
     }
```

altayhunoglu (approved):
Thank you for the nice explanation and solution Erdem.

Issue #221, opened 2021-05-15
1216. Valid Palindrome III


altayhunoglu:
https://leetcode.com/problems/valid-palindrome-iii/

Varsayılan kod:
```

class Solution {
    public boolean isValidPalindrome(String s, int k) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380538-4258f400-b5eb-11eb-8b19-8a30e8719cbd.png)
![image](https://user-images.githubusercontent.com/60903744/118380545-47b63e80-b5eb-11eb-867c-97d39ee3e0ea.png)

PR #382, opened 2021-07-23
1216. Valid Palindrome III

ErdemT09:
Resolves: #221 

## Algorithm:
Being able to remove some characters basically means that the palindrome does not have to be contiguous, a substring. It can thus be a subsequence. Our task is then to find the longest palindromic subsequence, and check if it's longer than the string size - k removed characters.

We can solve this problem using dynamic programming. Notice the sort of recursiveness here: the maximum palindromic subsequence length (`mpsl`) at some index `i` can be the maximum mspl starting at some index `j`. Or, if characters at these indices are equal, 2+mspl starting at some index `k, i>k>j`, mspl between i and j. It's the maximum of these two. 

We can easily derive a dp procedure from this. We pick some index and look for all indices before it. Initially the mspl-between is 0. And all mspl's for each index is 1. If two characters are equal, we increment the previous characters mspl to mspl-between+2. We check the global max each time. mspl-between is then updated to the previous value of the mspl of that previous index, we save it too in case it gets changed.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/ValidPalindromeIII.java
+++ b/src/main/java/algorithms/curated170/hard/ValidPalindromeIII.java
@@ -1,22 +1,37 @@
 package algorithms.curated170.hard;
 
-public class HandshakesThatDontCross {
+import java.util.Arrays;
+
+public class ValidPalindromeIII {
+    
     public boolean isValidPalindrome(String s, int k) {
+        int len = s.length();
         char[] data = s.toCharArray();
-        int[] dp = new int[data.length];
-        Arrays.fill(dp, 1);
-        int res = 1;
-        for (int i = 0; i < dp.length; i++ ) {
-            int maxSoFar = 0;
+        int[] maxPalindromicSubsequences = new int[len];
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/hard/ValidPalindromeIII.java
```diff
@@ -0,0 +1,37 @@
...
+public class ValidPalindromeIII {
+    
+    public boolean isValidPalindrome(String s, int k) {
+        int len = s.length();
+        char[] data = s.toCharArray();
+        int[] maxPalindromicSubsequences = new int[len];
+        Arrays.fill(maxPalindromicSubsequences, 1);
+        int maxPalindrome = 1;
+
+        for (int i = 0; i < len; i++) {
```
altayhunoglu, line 14:
```suggestion
        for (int i = 1; i < len; i++) {
```

altayhunoglu (approved):
Thank you for teaching me this elegant solution Erdem.
Approved.

Issue #225, opened 2021-05-15
642. Design Search Autocomplete System


altayhunoglu:
https://leetcode.com/problems/design-search-autocomplete-system/

Varsayılan kod:

```
class AutocompleteSystem {

    public AutocompleteSystem(String[] sentences, int[] times) {
        
    }
    
    public List<String> input(char c) {
        
    }
}

/**
 * Your AutocompleteSystem object will be instantiated and called as such:
 * AutocompleteSystem obj = new AutocompleteSystem(sentences, times);
 * List<String> param_1 = obj.input(c);
 */
```
![image](https://user-images.githubusercontent.com/60903744/118380627-dc20a100-b5eb-11eb-906b-bc04de974316.png)
![image](https://user-images.githubusercontent.com/60903744/118380630-e347af00-b5eb-11eb-99af-4184653562fc.png)
![image](https://user-images.githubusercontent.com/60903744/118380635-e80c6300-b5eb-11eb-9c35-c20b31da29fe.png)
![image](https://user-images.githubusercontent.com/60903744/118380638-ee9ada80-b5eb-11eb-8dba-458e48f9a460.png)
![image](https://user-images.githubusercontent.com/60903744/118380642-f5c1e880-b5eb-11eb-9b97-86f816effa85.png)

PR #383, opened 2021-07-23
642. Design Search Autocomplete System

ErdemT09:
Resolves: #225 

## Algorithm:
Let's consider what we are doing with the input here. After we are given some char, we only return the top three of the new char if the new char appears in these top three after the previous character. This naturally corresponds to Trie like structure, where characters have children going down the tree. 

What we should to is then create a Trie from the input sentences. This is pretty easy to implement. For the frequencies part, we first place them into a map. While creating the Trie, we also keep the top three words for each trie node too, inserting if frequencies are suitable. We keep this list sorted for convenient insertion, as it only has at most 3 elements. 

For the input we receive, we simply go down the trie, returning the "topThree" list. When we receive the termination input wirh `'#'`, we add the previous characters to the trie as we had done initially with the input sentences, perhaps updating the top three lists on the way.


altayhunoglu (approved):
A very nice implementation of Trie data structure.
Good explanation and solution.
Nothing further to add on my side.
Thank you, Erdem.

Issue #229, opened 2021-05-15
772. Basic Calculator III


altayhunoglu:
https://leetcode.com/problems/basic-calculator-iii/

Varsayılan kod:

```
class Solution {
    public int calculate(String s) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380709-87c9f100-b5ec-11eb-8222-1e6f33035f34.png)
![image](https://user-images.githubusercontent.com/60903744/118380712-8e586880-b5ec-11eb-85d6-04f9d5a7d196.png)
![image](https://user-images.githubusercontent.com/60903744/118380716-931d1c80-b5ec-11eb-9ddc-d7871e701a04.png)

PR #384, opened 2021-07-23
772. Basic Calculator III

ErdemT09:
Resolves: #229 

## Algorithm: 
What we have to do here is parsing some given string into a numerical calculation. 
Here are few catches:

- Plus and minus signs are like check points.

- The operator before a number determines what should be done with it to its previous sum:

`100-40/2/5` ->
40 gets subtracted from the previous sum 0 -> -40 
Then previous sum divided by 2 yields -> -20
Same for 5 -> -4
At the end, it is added to the total sum, which was 100 -> 100+(-4)=96

- For parentheses we call to function again, in a case like this:

`calculate(e1*(e2+(e3)))`
What we should do is:
```
calculate(e1)*calculate(e2+(e3)) ->
calculate(e1)*(calculate(e2) + calculate(e3))
```
Which is pretty easy to implement using recursion, with closing parentheses terminating the recursion.

So, we need to keep track of the total sum, current operator and the current previous sum to implement these.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/BasicCalculatorIII.java
+++ b/src/main/java/algorithms/curated170/hard/BasicCalculatorIII.java
@@ -1,0 +1,52 @@
+package algorithms.curated170.hard;
+
+public class BasicCalculatorIII {
+
+    private int i = 0;
+
+    public int calculate(String s) {
+        int n = s.length();
+        int num = 0;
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Thank you, Erdem. Elegant solution and nice explanation.

Issue #222, opened 2021-05-15
471. Encode String with Shortest Length


altayhunoglu:
https://leetcode.com/problems/encode-string-with-shortest-length/

Varsayılan kod:

```
class Solution {
    public String encode(String s) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380565-63214980-b5eb-11eb-8106-8a453b656b2d.png)
![image](https://user-images.githubusercontent.com/60903744/118380568-69afc100-b5eb-11eb-96cd-36ad7993a8bb.png)
![image](https://user-images.githubusercontent.com/60903744/118380570-70d6cf00-b5eb-11eb-83ae-8cba71ae5403.png)

PR #385, opened 2021-07-23
394. Decode String

ErdemT09:
Problem: https://leetcode.com/problems/decode-string/

This problem is like the inverse of #222, so it is important. It also helps practicing parsing.

## Algorithm: 
As long as we don't see any numbers, we add the current characters to our result, we don't need to check for brackets because they only come after numbers. When we see a digit, we try to get its full value, as it can consist of more than a single digit. After that we increment the index to skip the bracket, then recursively call the function to parse the inner of the bracket. The parsing loop then should stop when it encounters a closing bracket, or it reaches the end of the string. 

Once we have read inner of some brackets, we increment the index to skip the closing brackets. We add it to the result as many times as we have previously read before the brackets, thus adding it as if it was multiplied.

### DIFF
--- a/src/main/java/algorithms/medium/DecodeString.java
+++ b/src/main/java/algorithms/medium/DecodeString.java
@@ -2,28 +2,41 @@ package algorithms.medium;
 
 public class DecodeString {
 
-    int index = 0;
+    int i = 0;
+    char[] data;
+    int len;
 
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Very elegant solution.
Thank you, Erdem.

PR #386, opened 2021-07-25
471. Encode String with Shortest Length

ErdemT09:
Resolves: #222 

Draft

### DIFF
--- a/src/main/java/algorithms/curated170/hard/EncodeStringWithShortestLength.java
+++ b/src/main/java/algorithms/curated170/hard/EncodeStringWithShortestLength.java
@@ -3,43 +3,20 @@ package algorithms.curated170.hard;
 import java.util.HashMap;
 
 public class EncodeStringWithShortestLength {
-    HashMap<String, String> map;
+    HashMap<String, String> encodeMap;
 
     public String encode(String s) {
-        map = new HashMap<>();
+        encodeMap = new HashMap<>();
         return buildEncodedStr(s);
     }
 
     private String buildEncodedStr(String s) {
-        if (map.containsKey(s)) {
-            return map.get(s);
+        if (encodeMap.containsKey(s)) {
+            return encodeMap.get(s);
         }
         char[] data = s.toCharArray();
         int len = data.length;
-        String[][] ss = new String[len][len];
-
-        for (int i = 0; i < len; i++) {
-            int[] preStartIdx = new int[len - i + 1];
-            preStartIdx[0] = -1;
-            int repHeadLen = -1;
-
-            for (int j = 0; i + j < len; j++) {
-                repHeadLen = findRepetitionStart(data, i, preStartIdx, repHeadLen, j);
-                
-                int totalCodeLen = j + 1;
-                preStartIdx[totalCodeLen] = repHeadLen++;
-
-                int encodedStrLen = hasMultipleOfSubstr(repHeadLen, totalCodeLen) ? totalCodeLen - repHeadLen
-                        : totalCodeLen;
-                        
-                if (encodedStrLen == totalCodeLen) {
-                    ss[i][i + totalCodeLen - 1] = s.substring(i, i + totalCodeLen);
-                } else {
-                    String s1 = deliverEncodedStr(s, i, totalCodeLen, encodedStrLen);
-                    ss[i][i + j] = s1.length() < totalCodeLen ? s1 : s.substring(i, i + totalCodeLen);
-                }
-            }
-        }
+        String[][] ss = produce(s, data, len);
 
         int[] dp = new int[len + 1];
         int[] prev = new int[len + 1];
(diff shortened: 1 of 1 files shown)


src/main/java/algorithms/curated170/hard/EncodeStringWithShortestLengthDP.java
```diff
@@ -0,0 +1,53 @@
+package algorithms.curated170.hard;
+
+public class EncodeStringWithShortestLengthDP {
```
altayhunoglu, line 3: We may find some ways to simplify this code, I think. It is hard to fallow.

altayhunoglu:
Note for the KMP solution:

https://youtu.be/V5-7GzOfADQ

src/main/java/algorithms/curated170/hard/EncodeStringWithShortestLengthDP.java
ErdemT09: Is it more clear now?

src/main/java/algorithms/curated170/hard/EncodeStringWithShortestLengthDP.java
```diff
@@ -0,0 +1,56 @@
...
+
+            initializeEndings(s, codings, e);
+
+            for (int b = e - 1; b - (e - b) + 1 >= 0; --b) {
+                String pattern = s.substring(b + 1, e + 1);
+                int patternLen = e - b;
+
+                int maxRepStartIdx = b - patternLen + 1;
+                for (int repStart = maxRepStartIdx; repStart >= 0
+                        && pattern.equals(s.substring(repStart, repStart + patternLen)); repStart -= patternLen) {
```
altayhunoglu, line 19: Thanks, Erdem.
It is remarkably cleaner now.
Let us lighten the burden of for-loop to make it simpler:
```suggestion
    for (int repStart = maxRepStartIdx; repStart >= 0; repStart -= patternLen) {
                   if(!pattern.equals(s.substring(repStart, repStart + patternLen)))
                    break;
```

altayhunoglu (approved):
Thank you, Erdem. It is a hard question but you handled it quite fine.

ErdemT09:
> Thank you, Erdem. It is a hard question but you handled it quite fine.

ErdemT09:
### Approach 2-DP with Updates

Similar to the dynamic programming approach above, we can group the best codings of `s` by their end and their start.
So, let the return value `best[0][length-1]`.

For each index end index `e`, best[e][e] is equal to `s[e]`.
Our initial best guess for `best[i][e]` is then `best[i][e-1] + best[e][e]`, `i<e`.

Here, there are possible patterns that can occur and end at `e`.
These patterns start at some index `b+1`. They thus have the length `e-b`. In order to check if this pattern occurs before, we need to start by checking substrings that start at index `r = b+1-patternLength`. First we compare the substrings `s[r:r+patternLength]` and `s[b+1:e+1]`. If they are equal, this means that we can make an encoding in the form `"2[ pattern ]"`. But this pattern itself might have been encoded. So, we refer to our dp array for its coding -> `"2[ best[b+1][e] ]"`. If this encoding is shorter than `best[r][e]`, we update it. But this would also imply that we need to change our inital values. So, we update:
`best[u][e]` = `best[u][r-1] + encodedStr`, `u<r`.

After this, we can decrement `r` `patternLength` indices back. We then check for `"3[ best[b+1][e] ]"`, then `"4[ best[b+1][e] ]"` etc.,
Incrementally, we build up the way to `best[0][length-1]`.

ErdemT09:
Here is approach 3. It is a dp approach like the other 3, but its simpler and slower. I'm leaving it for the future.

```java
class Solution {
    Map<Integer, Map<Integer, String>> map = new HashMap<Integer, Map<Integer, String>>();
    public String encode(String s) {
        return minimize(s, 0, s.length()-1);
    }

    public String minimize(String s, int i, int j){
        if(map.containsKey(i) && map.get(i).containsKey(j)) return map.get(i).get(j);

        // first see if it can be repeated several times
        int minLen = j - i + 1;
        String res = s.substring(i, j+1);
        for(int l = 1; l <= (j-i+1)/2; l++) {
            if( (j-i+1) % l != 0) continue;

            boolean ok = true;
            for(int k = 0; k < (j - i + 1); k++) {
                if(s.charAt(i+k) != s.charAt(i + k%l)) {
                    ok = false;
                    break;
                }
            }
            if(ok) {
                String assemble = (j-i+1)/l + "[" + minimize(s, i, i+l-1) + "]";
                if(assemble.length() < minLen) {
                    minLen = assemble.length();
                    res = assemble;
                }
            }
        }

        for(int k = i; k < j; k++) {
            String mix = minimize(s, i, k) + minimize(s, k+1, j);
            if(mix.length() < minLen) {
                minLen = mix.length();
                res = mix;
            }
        }

        if(!map.containsKey(i)) map.put(i, new HashMap<Integer, String>());
        map.get(i).put(j, res);
        return res;
    }

}
```

altayhunoglu (approved):
Thanks again, Erdem.
When we turn back, we will find a way to visualize this kind of complex questions.

Issue #226, opened 2021-05-15
588. Design In-Memory File System


altayhunoglu:
https://leetcode.com/problems/design-in-memory-file-system/

Varsayılan kod:

```
class FileSystem {

    public FileSystem() {
        
    }
    
    public List<String> ls(String path) {
        
    }
    
    public void mkdir(String path) {
        
    }
    
    public void addContentToFile(String filePath, String content) {
        
    }
    
    public String readContentFromFile(String filePath) {
        
    }
}

/**
 * Your FileSystem object will be instantiated and called as such:
 * FileSystem obj = new FileSystem();
 * List<String> param_1 = obj.ls(path);
 * obj.mkdir(path);
 * obj.addContentToFile(filePath,content);
 * String param_4 = obj.readContentFromFile(filePath);
 */
```
![image](https://user-images.githubusercontent.com/60903744/118380651-11c58a00-b5ec-11eb-9abb-6e7316301a60.png)
![image](https://user-images.githubusercontent.com/60903744/118380656-18540180-b5ec-11eb-958c-a6502132561b.png)
![image](https://user-images.githubusercontent.com/60903744/118380659-1e49e280-b5ec-11eb-9087-ab2ff6b5d4a5.png)
![image](https://user-images.githubusercontent.com/60903744/118380662-23a72d00-b5ec-11eb-845f-20351819b874.png)

PR #387, opened 2021-07-25
588. Design In-Memory File System

ErdemT09:
Resolves: #226 

I recommend checking out #188 before this problem, it's kind of like the simpler version.

## Algorithm:
Each file/directory here has few properties:
```
- name
- child directories and files as a map (for directories)
- content (for files)
- is file = (content = null)
```
We can use this to create a general class for both folders and directories.

With this, each method reduces down to simple operations:

- `ls()`:
Find the specified folder or file.
If directory, return the key set of the child directories and files, sorted.
If file, return name.
- `mkdir()`:
Find the specified folder or file, create new directories on the way if needed.
- `addContentToFile()`:
Find the specified file, add to its content.
- `readContentFromFile()`:
Find the specified file, get its content.

The common problem here is finding the file or the directory. Hence, it's only most convenient to create a separate method for it. We simply get the path name and split it into strings originally separated by '/'. From the root directory we are keeping, we simply traverse down to the sought folder using the maps of the directories. We also create new directories when we can't find the directory name we had in the paths string.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/DesignInMemoryFileSystem.java
+++ b/src/main/java/algorithms/curated170/hard/DesignInMemoryFileSystem.java
@@ -1,0 +1,90 @@
+package algorithms.curated170.hard;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.TreeMap;
+
+public class DesignInMemoryFileSystem {
+
+    class FileSystem {
+
+        public Directory root;
+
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Quite clear and self-expressive code.
Nothing to add on my side.
Thanks Erdem.

Issue #239, opened 2021-05-15
1153. String Transforms Into Another String


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public boolean canConvert(String str1, String str2) {
        
    }
}
```

![image](https://user-images.githubusercontent.com/60903744/118380927-839ed300-b5ee-11eb-9f38-495748aead2f.png)
![image](https://user-images.githubusercontent.com/60903744/118380929-8994b400-b5ee-11eb-9571-03fb241258a1.png)

PR #388, opened 2021-07-25
1153. String Transforms Into Another String

ErdemT09:
https://leetcode.com/problems/string-transforms-into-another-string/

Resolves: #239 

## Algorithm:
A string is not convertible to another string under these 2 conditions:
1. A character transforms into more than one character:
`"aaa" -> "hgf"` is not possible.
2. The second string contains all possible letters:
When both strings contain same letters as in: `"abc" -> "bca"`, we can convert individual letters without making them same by temporarily converting them to another letters like this:
`"abc" -> "dbc" -> "dba" -> "dca" -> "bca"`
As long as we have a letter not present in the destination string, we can do such temporary conversions.

We only have to check for these.


altayhunoglu (approved):
Nice code and it is good to see that our names can be converted:

Thank you Erdem.

image

Issue #242, opened 2021-05-15
631. Design Excel Sum Formula


altayhunoglu:
https://leetcode.com/problems/design-excel-sum-formula/

Varsayılan kod:

```
class Excel {

    public Excel(int height, char width) {
        
    }
    
    public void set(int row, char column, int val) {
        
    }
    
    public int get(int row, char column) {
        
    }
    
    public int sum(int row, char column, String[] numbers) {
        
    }
}

/**
 * Your Excel object will be instantiated and called as such:
 * Excel obj = new Excel(height, width);
 * obj.set(row,column,val);
 * int param_2 = obj.get(row,column);
 * int param_3 = obj.sum(row,column,numbers);
 */
```
![image](https://user-images.githubusercontent.com/60903744/118380994-0cb60a00-b5ef-11eb-90ef-2f4e623dd7fd.png)
![image](https://user-images.githubusercontent.com/60903744/118380997-18093580-b5ef-11eb-8a5d-4e5de25aba9b.png)

```
Example 1:

Input
["Excel", "set", "sum", "set", "get"]
[[3, "C"], [1, "A", 2], [3, "C", ["A1", "A1:B2"]], [2, "B", 2], [3, "C"]]
Output
[null, null, 4, null, 6]

Explanation
Excel excel = new Excel(3, "C");
 // construct a 3*3 2D array with all zero.
 //   A B C
 // 1 0 0 0
 // 2 0 0 0
 // 3 0 0 0
excel.set(1, "A", 2);
 // set mat[2]["B"] to be 2.
 //   A B C
 // 1 2 0 0
 // 2 0 0 0
 // 3 0 0 0
excel.sum(3, "C", ["A1", "A1:B2"]); // return 4
 // set mat[3]["C"] to be the sum of value at mat[1]["A"] and the values sum of the rectangle range whose top-left cell is mat[1]["A"] and bottom-right cell is mat[2]["B"].
 //   A B C
 // 1 2 0 0
 // 2 0 0 0
 // 3 0 0 4
excel.set(2, "B", 2);
 // set mat[2]["B"] to be 2. Note mat[3]["C"] should also be changed.
 //   A B C
 // 1 2 0 0
 // 2 0 2 0
 // 3 0 0 6
excel.get(3, "C"); // return 6
 
```
![image](https://user-images.githubusercontent.com/60903744/118381004-2d7e5f80-b5ef-11eb-96a4-4c3d8d19475c.png)

[...]

PR #389, opened 2021-07-25
631. Design Excel Sum Formula

ErdemT09:
Resolves: #242 

## Algorithm:
When calculating the value of each cell, there are 2 possibilities:
1. The cell only has a value.
2. The cell points to a collection of cells and/or rectangles.

`set()` method updates the value of some cell to an integer.
`sum()` method updates the value of some cell to a collection of cell names.

In order to get the value of summed cells, 2. type of cells, we try to calculate the value of each cell/rectangle they point at.
Let's say: `D4 = [B1:C2, A2]`. In order to calculate `D4`, we need to calculate The range `B1:C2`. This range might also contain references to other ranges, for example a case might be: `B2 = [A1:A3]`. For calculating this, we also need to sum this range. This yields a recursive method where we calculate the value of each cell each time by calculating the ranges they point to. When values are updated, the structure recalculates such values each time again, so we don't have to worry about updating something like the prefix sum etc. 

In implementation, cells are `Object`s that are either `Integer`, pure value, or `String[]`, reference to other cells by name, which gets parsed by the `calculate()` method.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/DesignExcelSumFormula.java
+++ b/src/main/java/algorithms/curated170/hard/DesignExcelSumFormula.java
@@ -5,67 +5,83 @@ import java.util.Arrays;
 public class DesignExcelSumFormula {
 
     class Excel {
+        
+        private static final int NO_POS = -1;
         Object[][] matrix;
-        int m, n;
+        private final int M, N;
 
         public Excel(int height, char width) {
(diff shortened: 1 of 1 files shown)


altayhunoglu (approved):
Thanks for the contribution Erdem.
Quite clear code and explanation.

Issue #230, opened 2021-05-15
317. Shortest Distance from All Buildings


altayhunoglu:
Varsayılan kod:

```
class Solution {
    public int shortestDistance(int[][] grid) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380722-ad56fa80-b5ec-11eb-89af-4bd13e5acecc.png)
![image](https://user-images.githubusercontent.com/60903744/118380724-b5af3580-b5ec-11eb-9d55-0c1695f8f644.png)
![image](https://user-images.githubusercontent.com/60903744/118380730-bcd64380-b5ec-11eb-9468-d3f032fe2318.png)

PR #390, opened 2021-07-26
317. Shortest Distance from All Buildings

ErdemT09:
Resolves: #230 

## Algorithm:
This is a pretty simple grid-BFS problem. To find the tile with minimum distance, we need to count the distances to tiles. We can count these distances by BFS'ing the whole grid at each building and summing up the tile distances in some matrix. Let'say that we have visited some tiles for some building. In some another building, if any tile around it can visit hasn't been visited by all the buildings before it, that means that there is no shortest distance for all buildings. This can be implemented by adding a check to the BFS, where this visit count property can hinder the process. We can also use the given space to count the visits in the given grid matrix. Eventually, we return the least distance we encountered for the last building.


altayhunoglu (approved):
Thank you, Erdem. Nothing further to add from my side.
It is quite clear and self-expressive code.

Issue #236, opened 2021-05-15
308. Range Sum Query 2D - Mutable


altayhunoglu:
https://leetcode.com/problems/range-sum-query-2d-mutable/

Varsayılan kod:

```
class NumMatrix {

    public NumMatrix(int[][] matrix) {
        
    }
    
    public void update(int row, int col, int val) {
        
    }
    
    public int sumRegion(int row1, int col1, int row2, int col2) {
        
    }
}

/**
 * Your NumMatrix object will be instantiated and called as such:
 * NumMatrix obj = new NumMatrix(matrix);
 * obj.update(row,col,val);
 * int param_2 = obj.sumRegion(row1,col1,row2,col2);
 */
```

![image](https://user-images.githubusercontent.com/60903744/118380857-dc21a080-b5ed-11eb-8023-b23a20a31663.png)
![summut-grid](https://user-images.githubusercontent.com/60903744/118380861-e774cc00-b5ed-11eb-800c-84ec1a77846a.jpg)
![image](https://user-images.githubusercontent.com/60903744/118380864-ecd21680-b5ed-11eb-8619-6906c86fff05.png)
![image](https://user-images.githubusercontent.com/60903744/118380865-f0fe3400-b5ed-11eb-8a4c-47a05860aef6.png)

PR #392, opened 2021-07-28
307. Range Sum Query - Mutable & 308. Range Sum Query 2D - Mutable

ErdemT09:
Problem: https://leetcode.com/problems/range-sum-query-mutable/

Here, the problem we face when we try to sum ranges using brute force range summing is that we cannot use prefix sums. If the range was immutable, it would have been a well choice. But reacting to the updates with prefix sums itself would have been inefficient. There are few ways of resolving this inconsistency between partial array sums and update/sum queries:

### Approach 1-Segment Trees
We could think of the range sums as a tree to some degree.
The root of the tree has the value `sum[0:length-1] = sum[0:mid]+sum[mid+1:length-1]`. Thus, each node in this range tree has a value equal to the sum of its 2 children. 
For the base case in the leaves of the tree, the value would be equal to `sum[i:i] = arr[i]`. Above the base case, for example at the node with the range 0:1, the value would be equal to the sum of the values of the children `= sum[0:0] + sum[1:1]`. Thus, after calculating the values of a node's children, we can calculate the nodes value. 

We can naturally implement this using custom classes and such, but that is inefficient. A more efficient implementation can be done through arrays.

- #### Build
Let's say that the root of the tree is located at the index 1. Then we can create a level-order tree where the next 2 indices are the children of the root at index 1, i.e. 2=left child, 3=right child. From this, we can infer that the second half of the tree array will consist of the base case values of individual indices. After defining these values, to continue building up in a bottom-up manner, we start at the index before the half, and the values as `tree[i] = tree[2*i] + tree[2*+1], 2*i = left child, 2*i+1 right child`. 
For `tree[]`, we naturally need an array of size `2*n` (n=array length).

- #### Range Sum Queries
Let `i` be the left index of the query, `j` the right.
Initially, we must add `n` to both i and j because their base cases are located `n` indices after their original array locations.

From the building of the tree array, we know that left children are located at even indices and right children at odds. 
This has two implications for i and j:

-- For `i`: If i is an even number, that means that we need to add the sum of its parent, not just the individual node's. If i is an odd number, that means that the value of the left child should be excluded, we only add the right child's value.
In the 1. case, we simply shift i to its parent index by dividing it by 2. In the 2. case, we add the right child's value. We then increment i. This would get us to the left node of the right sibling of the parent, whose value could be entirely add if `i≤j`.

-- For `j`: Likewise, if j is an odd number, that means that the node is a right child and the parent's range is included to be added. If j is an even number, this implies that the right child's value needs to be excluded. The procedure is symmetrical to the one done for `i`:
In the 1. case, we shift j to the parent index by dividing by 2 (integer division). In the 2. case, we add the left child's value. We then decrement i. This would get us to the right node of the left sibling of the parent, whose value could be entirely add if `i≤j`.

We do these 2 procedure together.

- #### Update Queries
This is pretty simple too. We first set tree[i+n] to the value, then update its parents values: tree[i/2] = tree[i] + tree[i^1]. The xor operation is a shortcut for using the left child if i is the right, or the right child if i is the left.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/rangesumquery2dmutable/RangeSumQuery2DMutable.java
+++ b/src/main/java/algorithms/curated170/hard/rangesumquery2dmutable/RangeSumQuery2DMutable.java
@@ -1,0 +1,11 @@
+package algorithms.curated170.hard.rangesumquery2dmutable;
+
+public class RangeSumQuery2DMutable {
+
+
+    
+
+    public static void main(String[] args) {
+
+    }
+}
--- a/src/main/java/algorithms/curated170/hard/rangesumquery2dmutable/RangeSumQuery2DMutable.java
+++ b/src/main/java/algorithms/curated170/hard/rangesumquery2dmutable/RangeSumQuery2DMutable.java
@@ -2,10 +2,43 @@ package algorithms.curated170.hard.rangesumquery2dmutable;
 
 public class RangeSumQuery2DMutable {
 
+    class NumMatrix {
+        
+        private int[][] colSums;
+        private int[][] matrix;
 
-    
+        public NumMatrix(int[][] matrix) {
+            this.matrix = matrix;
+            
+            int m = matrix.length;
+            int n = matrix[0].length;
+            colSums = new int[m + 1][n];
+            
+            for (int i = 1; i <= m; i++) {
+                for (int j = 0; j < n; j++) {
+                    colSums[i][j] = colSums[i - 1][j] + matrix[i - 1][j];
+                }
+            }
+        }
 
-    public static void main(String[] args) {
+        public void update(int row, int col, int val) {
+            for (int i = row + 1; i < colSums.length; i++) {
+                colSums[i][col] += (val - matrix[row][col]);
+            }
 
+            matrix[row][col] = val;
+        }
+
+
+        public int sumRegion(int row1, int col1, int row2, int col2) {
+            int sum = 0;
+
+            for (int j = col1; j <= col2; j++) {
+                sum += colSums[row2 + 1][j] - colSums[row1][j];
+            }
+
+            return sum;
+        }
     }
+
 }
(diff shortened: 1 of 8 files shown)


ErdemT09:
### Approach 2-Binary Indexed Tree (Finweck Tree)

ErdemT09:
### Approach 3-Half Counting Solution
When summing a range, if the length of the range is larger than half of the array's, then we can subtract values out of the range from the sum of the whole array instead of counting the range itself. This seems pretty slow in theory. However, it is fast fast in implementation.

ErdemT09:
## 304. Range Sum Query 2D - Immutable
Problem: https://leetcode.com/problems/range-sum-query-2d-immutable/

Here, we use the same matrix prefix calculation algorithm we used for #358. It's pretty simple.

ErdemT09:
## 308. Range Sum Query 2D - Mutable
Resolves: #236

altayhunoglu:
The essence of all these solutions is "modularization" (and our container idea) indeed.

We divide the array into multiple segments to prevent the states of the others parts are effected just because the update of an independent part.

altayhunoglu:
The essence of all these solutions is "modularization" (and our container idea) indeed.

We divide the array into multiple segments to prevent the states of the other segments being affected just because an independent part is updated.

ErdemT09:
Further notes for Segment and Fenwick Trees:
We can also add other usage of these trees. Namely, min/max finding, greatest common divisor finding, range/region updates, zero counting etc.

More resources:
https://cp-algorithms.com/data_structures/segment_tree.html
https://leetcode.com/articles/a-recursive-approach-to-segment-trees-range-sum-queries-lazy-propagation/
https://cp-algorithms.com/data_structures/fenwick.html
https://www.topcoder.com/thrive/articles/Binary%20Indexed%20Trees
https://github.com/williamfiset/Algorithms/tree/master/src/main/java/com/williamfiset/algorithms/datastructures/fenwicktree

altayhunoglu (approved):
I could not see anything further to add on my side.
Your valuable efforts are much appreciated Erdem.
Thank you.

Issue #232, opened 2021-05-15
727. Minimum Window Subsequence


altayhunoglu:
https://leetcode.com/problems/minimum-window-subsequence/

Varsayılan kod:

```
class Solution {
    public String minWindow(String s1, String s2) {
        
    }
}
```
![image](https://user-images.githubusercontent.com/60903744/118380770-10e12800-b5ed-11eb-9fb4-e28692205bf8.png)
![image](https://user-images.githubusercontent.com/60903744/118380775-1c345380-b5ed-11eb-8a49-d675d74fad76.png)

PR #393, opened 2021-07-29
727. Minimum Window Subsequence

ErdemT09:
Resolves: #232 

### Approach 1-Character Finding
Let's say that we are checking for the subsequence "abcd" in the string "abcabcd".
First we incrementally look for indices of a,b,c,d. We get to the end of the string. In order to make sure that this is the minimum window between the start and the end, we must also decrementally look for characters of the target string that have appeared before the end index. If we only had the string "abcwwwd" to search in, the only previous occurrence of the target would have indeed been the entire string. However, we need to go back to find the character c starting before, b starting before, a starting before etc. 

### Approach 2-Dynamic Programming
Over the range of the String S, we can look for closest subsequence starts. Then, for all occurrences of the second character of the string T in S, we look for the closest first indices. For the third character, closest second character index before, which themselves point out to the subsequence start. With this incremental structure, we find the closest subsequence starts for the last character of the target string. We look for the minimum and return the corresponding substring. It's to be noted that this approach is slower, although worth learning about.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/MinimumWindowSubsequence.java
+++ b/src/main/java/algorithms/curated170/hard/MinimumWindowSubsequence.java
@@ -1,0 +1,31 @@
+package algorithms.curated170.hard;
+
+public class MinimumWindowSubsequence {
+
+    private static final int NO_POS = -1;
+
+    public String minWindow(String S, String T) {
+        int i = NO_POS;
+        String minWin = "";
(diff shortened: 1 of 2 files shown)


altayhunoglu (approved):
Thanks, Erdem. Nothing to add on my side.
I tried to narrow down the source string for the DP solution but no effect. We can leave for now.

Issue #210, opened 2021-05-15
428. Serialize and Deserialize N-ary Tree


altayhunoglu:
https://leetcode.com/problems/serialize-and-deserialize-n-ary-tree/

Varsayılan kod:
```
/*
// Definition for a Node.
class Node {
    public int val;
    public List<Node> children;

    public Node() {}

    public Node(int _val) {
        val = _val;
    }

    public Node(int _val, List<Node> _children) {
        val = _val;
        children = _children;
    }
};
*/

class Codec {
    // Encodes a tree to a single string.
    public String serialize(Node root) {
        
    }
	
    // Decodes your encoded data to tree.
    public Node deserialize(String data) {
        
    }
}

// Your Codec object will be instantiated and called as such:
// Codec codec = new Codec();
// codec.deserialize(codec.serialize(root));
```

![image](https://user-images.githubusercontent.com/60903744/118380172-7d0d5d00-b5e8-11eb-9456-28c66b3f656b.png)
![image](https://user-images.githubusercontent.com/60903744/118380175-8696c500-b5e8-11eb-894b-4329a40bed88.png)
![image](https://user-images.githubusercontent.com/60903744/118380182-8dbdd300-b5e8-11eb-954c-5ea7a65f0824.png)

PR #394, opened 2021-07-31
428. Serialize and Deserialize N-ary Tree

ErdemT09:
Resolves: #210 

Firstly, this problem has quite a many possible solutions. There are many creative ways of packing a data structure like a tree into a string form. However, here, I am only going to be considering ones that are the most useful, to learn about or to actually use.

### Approach 1-Nested Lists of Children
By using the recursive definition of the trees, we can encode each tree in the following form:
```
encode(T):
if T = null: ""
else: T.val+"["+encode(every child of T)+"]"
```
We can thus encode trees into relatively readable structures like: `9[27[0[]18[9[]]]54[]63[999[]]]`

For the decoding, we simply check if there are characters, node values that we can read. If so, we add a node to our nodes stack. After that, if there are new integer values, that means that there are new nodes and we similarly add them. We don't encounter any integer values, that means that the last node in the stack has come to an end. Then we pop it and add it as a child to the previous node in the stack. If popping it makes the stack empty, it implies that this the node that we return as the root.

We can also implement this using recursion, it's however a bit more complex in Java because of passing by reference/value issues.

### Approach 2-Level Order Encoding with Extended ASCII/Unicode Characters as Child Count
To denote that the children of a node has come to an end, we can put a termination character such as '#'.
A node with value 9 with children 999, 18 and 36 would then look like this:
9999#18#36##
The problem is that the values of root and the first children are concatenated if we add them like this. Instead of this, we can add them as character values.
Instead of 9, we would add '0'=48(>ASCII code  of '#') to it. Same for 999. Then, the encoding would look like this but as a string:
`[\u0057, \u1047, '#', \u0066` and so on

For decoding, we basically treat all the characters up to some termination point as the node's children. Since the nodes themselves would also look for their termination points, the end of the root won't mix up with the end of their children.

### Approach 3-Level Order Encoding with Child Count + DFS Decoding
We can decode each node with the number of their children:
A node with 4 children can be encoded as:
`root_val, 4, ...`
We can recursively encode the nodes of a tree like this. We do write '0' for leaf nodes too.

Decoding is quite easy here. We first split all the substring that were separated by commas and put them into a queue. The first value we poll is the root, second its children `count`. We recursively decode next `count` nodes as the children of this node. The recurrence takes care of preventing any confusions of polling and decoding.

### DIFF
--- a/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
+++ b/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
@@ -49,22 +49,6 @@ public class SerializeAndDeserializeNaryTreeChildCount {
         }
     }
 
-    static class Node {
-        public int val;
-        public List<Node> children = new ArrayList<>();
-
-        public Node() {
-        }
-
-        public Node(int _val) {
-            val = _val;
-        }
-
-        public Node(int _val, List<Node> _children) {
-            val = _val;
-            children = _children;
-        }
-    }
 
     public static void main(String[] args) {
         Node c1 = new Node(10);
--- a/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
+++ b/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
@@ -1,4 +1,4 @@
-package algorithms.curated170.hard;
+package algorithms.curated170.hard.serializeanddeserializenarytree;
 
 import java.util.ArrayList;
 import java.util.Arrays;
--- a/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
+++ b/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
@@ -55,7 +55,7 @@ public class SerializeAndDeserializeNaryTreeChildCount {
         Node c1 = new Node(10);
         Node c2 = new Node(77);
         Node c3 = new Node(24);
-        Node root = new Node(90, List.of(c1, c2, c3));
+        Node root = new Node(90,   new ArrayList<>(List.of(c1, c2, c3)));
 
         var codec = new Codec();
         String coded = codec.serialize(root);
--- a/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
+++ b/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
@@ -10,7 +10,6 @@ public class SerializeAndDeserializeNaryTreeChildCount {
 
     static class Codec {
 
-        // Encodes a tree to a single string.
         public String serialize(Node root) {
             List<String> list = new ArrayList<>();
             encode(root, list);
@@ -29,7 +28,6 @@ public class SerializeAndDeserializeNaryTreeChildCount {
             }
         }
 
-        // Decodes your encoded data to tree.
         public Node deserialize(String data) {
             if (data.isEmpty())
                 return null;
--- a/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
+++ b/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
@@ -5,6 +5,7 @@ import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
+import algorithms.curated170.medium.Node;
 
 public class SerializeAndDeserializeNaryTreeChildCount {
 
--- a/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeNestedList.java
+++ b/src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeNestedList.java
@@ -56,22 +56,7 @@ public class SerializeAndDeserializeNaryTreeNestedList {
         }
     }
 
-    static class Node {
-        public int val;
-        public List<Node> children = new ArrayList<>();
-
-        public Node() {
-        }
-
-        public Node(int _val) {
-            val = _val;
-        }
-
-        public Node(int _val, List<Node> _children) {
-            val = _val;
-            children = _children;
-        }
-    }
+ 
 
     public static void main(String[] args) {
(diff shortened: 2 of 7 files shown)


src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeNestedList.java
```diff
@@ -0,0 +1,86 @@
...
+
+        public Node(int _val) {
+            val = _val;
+        }
+
+        public Node(int _val, List<Node> _children) {
+            val = _val;
+            children = _children;
+        }
+    }
```
altayhunoglu, line 73:
```suggestion
 
```

We can use the existing one, I think.

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeNestedList.java
```diff
@@ -0,0 +1,86 @@
+package algorithms.curated170.hard.serializeanddeserializenarytree;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Stack;
+
```
altayhunoglu, line 6:
```suggestion
import algorithms.curated170.medium.Node;

```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
```diff
@@ -0,0 +1,80 @@
+package algorithms.curated170.hard.serializeanddeserializenarytree;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Queue;
+
```
altayhunoglu, line 8:
```suggestion
import algorithms.curated170.medium.Node;

```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeUnicode.java
```diff
@@ -0,0 +1,77 @@
+package algorithms.curated170.hard.serializeanddeserializenarytree;
+
+import java.util.ArrayList;
+import java.util.List;
+
```
altayhunoglu, line 5:
```suggestion
import algorithms.curated170.medium.Node;

```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeUnicode.java
```diff
@@ -0,0 +1,77 @@
...
+        public Node(int _val) {
+            val = _val;
+        }
+
+        public Node(int _val, List<Node> _children) {
+            val = _val;
+            children = _children;
+        }
+    }
```
altayhunoglu, line 64:
```suggestion

```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
```diff
@@ -0,0 +1,80 @@
...
+        public Node(int _val) {
+            val = _val;
+        }
+
+        public Node(int _val, List<Node> _children) {
+            val = _val;
+            children = _children;
+        }
+    }
```
altayhunoglu, line 67:
```suggestion
```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeChildCount.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public static void main(String[] args) {
+        Node c1 = new Node(10);
+        Node c2 = new Node(77);
+        Node c3 = new Node(24);
+        Node root = new Node(90, List.of(c1, c2, c3));
```
altayhunoglu, line 73:
```suggestion
        Node root = new Node(90,   new ArrayList<>(List.of(c1, c2, c3)));
```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeNestedList.java
```diff
@@ -0,0 +1,86 @@
...
+            val = _val;
+            children = _children;
+        }
+    }
+
+    public static void main(String[] args) {
+        Node c1 = new Node(10);
+        Node c2 = new Node(77);
+        Node c3 = new Node(24);
+        Node root = new Node(90, List.of(c1, c2, c3));
```
altayhunoglu, line 79:
```suggestion
        Node root = new Node(90,   new ArrayList<>(List.of(c1, c2, c3)));
```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeUnicode.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public static void main(String[] args) {
+        Node c1 = new Node(10);
+        Node c2 = new Node(77);
+        Node c3 = new Node(24);
+        Node root = new Node(90, List.of(c1, c2, c3));
```
altayhunoglu, line 70:
```suggestion
        Node root = new Node(90,   new ArrayList<>(List.of(c1, c2, c3)));
```

src/main/java/algorithms/curated170/hard/serializeanddeserializenarytree/SerializeAndDeserializeNaryTreeNestedList.java
ErdemT09: What's the difference?

altayhunoglu: It is just to be compatible with the existing Node class.
A better option would be changing the existing `Node` constructor method's parameter, using List instead of ArrayList.
It would be better because `List` is more abstract and abstraction is generally a good idea to be more adaptable and maintainable.
But this way was practical. You can choose to the better way instead.

ErdemT09: I think that it may perhaps be important in more applied programming, but it's fine here.

altayhunoglu: I mean it will give an error if we leave it like this.

altayhunoglu (approved):
Thank you for your valuable efforts Erdem.

altayhunoglu:
When the information encoded in `System 1` should be decoded in `System 2`, we should be able to reconstruct the information without any damage.

IMG-1708

